PR #4, opened 2016-06-29
Support for personalized exercises

Mankro:
Additionally, async file upload exercise may be configured to require only a subset of the listed files.


teemulehtinen:
Thanks for the number of new features that seem to work! However, I would like to question if the personal directories are required. Does exercise grading exist that needs information from previous grading times? For one, this will destroy the re-grading of submissions by definition. I thought it would be sufficient to solve the generated-version from uid and ordinal_number, and provide this for the grading preparation (as already implemented here).

Mankro:
@teemulehtinen I understand the concern about personal directories. Honestly, we are still a little unsure if we really need that kind of permanent storage of user-specific files that are available to the grader program. It depends on how the course assignments evolve when they are ported to the new system (mooc-grader). One course, which has used this kind of permanent file storage in the past, has graders that have an execution time of several hours and the grading was divided to two phases. The grader would read a marker file from the personal directory to decide which phase is active. This requires the teacher to do some manual work between the phases; ideally the grader could finish everything automatically but the implementation of the grader is not always trivial...

teemulehtinen:
Fine enough! Could you add a settings.py setting required to enable the personal directories features? This way we can be sure that no permanent disk space is required without setting that switch. Otherways this is good to go :+1:

Mankro:
@teemulehtinen Should the settings.py switch enable/disable all files related to personalized exercises, or just the use of user-specific personal directories? If a course uses any personalized exercise, then it must store the pregenerated exercise instances (files in a different directory than the personal directory). When personalized exercises are used, the personal directory stays empty unless the exercise configuration uses the `store_user_files` action.

util/templates.py
```diff
@@ -33,6 +37,12 @@ def render_configured_template(request, course, exercise, post_url, default=None
         template = default
     else:
         raise ConfigError("Missing \"template\" in exercise configuration.")
+    if "personalized" in exercise and exercise["personalized"]:
```
teemulehtinen: I think this preparation should be moved into _exercise_context as well (request can be added as parameter). Even better the _exercise_context could invoke a "context function" in personalized.py that sets up the required directories and context.

teemulehtinen:
My idea was that if the personal directories are NOT enabled in settings:
1. store_user_files will log error and stop
2. rendering exercise template should not create personal directories
3. prepare does not link personal directories

Running the manage.py command can still create the exercise instance directories.

This way the switch in settings will make the grader stateful, else it is stateless. The number of gradings or students has no effect on state, e.g. disk space required.

Mankro:
@teemulehtinen I have implemented and committed your suggestions. But there is a new downside: if `settings.ENABLE_PERSONALIZED_EXERCISES` is `False`, viewing a personalized exercise fails silently; the web page has no personalized content, which makes it rather broken. We could either raise an exception (using personalized exercise when the settings disable it) or check if pregenerated instances exist and proceed normally but without creating the user personal directory. This matches the idea that the manage.py command would still work even with `settings.ENABLE_PERSONALIZED_EXERCISES = False`. However, grading would not work since the grading needs the generated instance files, which are accessed through the personal directory link that prepare action may not create with the disabled setting. Which option do you prefer?

teemulehtinen:
Could we have cp_personal and cp_generated? The last one would get a path argument from the personalized.py select method. First one could output an error to feedback if switch is not enabled.

I'm sorry I have so many requests, and I hope you see some point in the stateless vs stateful ideology. I'm very pleased to have more skilled contributors, such as yourself, in this project!

util/personalized.py
```diff
@@ -0,0 +1,252 @@
...
+    pregenerated_dir = pregenerated_exercises_directory_path(course, exercise)
+    userids = sum(user_ids_from_string(userids_str))
+    instances = pregenerated_exercise_instances(course, exercise)
+    num_instances = len(instances)
+    if num_instances == 0:
+        raise access.config.ConfigError("Exercise is personalized but no exercise instances have been pregenerated")
+    
+    if "max_submissions_before_regeneration" in exercise:
+        # the generated exercise may be regenerated after submitting certain amount of times
+        instance = (userids + ((submission_number - 1) // exercise["max_submissions_before_regeneration"]) *
```
teemulehtinen: Yes, there are obvious problems with this. Possible solution:
instances = list(range(num_instances))
r = random.Random(userids)
r.shuffle(instances)
instance = instances[((submission_number - 1) // exercise["max_submissions_before_regeneration"]) % num_instances]

Mankro:
I have changed the regeneration instance select formula and added cp_generated to prepare action. The user directory no longer has any "generated" link and it only stores the personal files if `store_user_files` action is used. I updated the README and other comments accordingly. I changed the name of the settings switch since it only enables the personal directories; the pregenerated exercise instances can always be used in personalized exercises.

I don't mind implementing the changes. They have been sensible and a higher quality implementation is in the best interest of everyone.

PR #6, opened 2016-07-05
Add default template for displaying personalized content in the exercise description page

Mankro:
Additionally, add error logging in util/personalized.py if reading a generated/personal file fails.


Mankro:
I am a bit uncertain about the Django translations. Should I add a Finnish translation for the "Download" string in the template?

teemulehtinen:
Well I guess you knew that as you have committed translations before. It seems that our system does not know finnish translation for "Download" before.

      (moocgraderenv3)lehtint6@ws-162:~/git/mooc-grader$ ./manage.py shell
      Python 3.4.3 (default, Jun 30 2015, 10:16:14) 
      [GCC 4.6.3] on linux
      Type "help", "copyright", "credits" or "license" for more information.
      (InteractiveConsole)
      >>> from django.utils import translation
      >>> translation.activate('fi')
      >>> from django.utils.translation import ugettext as _
      >>> _("Download")
      'Download'

Mankro:
I had tried adding translations once before. I wasn't sure if the Finnish version would really be used at all, anyway I have added translations for the few strings I have created. I didn't know that the working copy shouldn't include any course configurations when running the commands (makemessages), but I don't think it caused any trouble last time (in June).

teemulehtinen:
Finnish version will kick in with index.yaml "lang: fi" or using get parameter "?lang=fi" for an exercise URL. The course configurations that include *.html/*.py may stop the makemessages in an error.

PR #7, opened 2016-07-07
Support for including template generated course configuration files

matiaslindgren:


teemulehtinen:
A useful and well formed feature :+1:

PR #15, opened 2017-12-14
Add active elements

roosapi:
Adds view type and templates for active elements. The view type acceptGeneralForm is effectively a combination of acceptPost and acceptFiles and should be able to handle general form submits that include files, text inputs and dropdowns (other input types not tested). The default template for the type currently creates a text input for any other input than file.


teemulehtinen:
The feature follows mooc-grader conventions and includes a new required view type. In future, we could deprecate some old view types that are becoming obsolete.

PR #19, opened 2018-03-16
Support feedback templates in container graded exercises

Mankro:
MOOC grader overrides the feedback coming from the container with the rendered template only if the exercise configuration has explicitly set a feedback template. When the template is used, the feedback HTML/text from the container can be included in the template by using variables.

Feedback templates were supported in the old chroot sandbox model and enabling them with container sandboxes could ease the transition. Furthermore, it is more convenient to write complicated HTML in templates than in grader programs (think of long HTML strings within Python code). The container tools/scripts do not yet provide a convenient default method for rendering complex templates inside the container, so this PR adds the possibility to use the feedback templates in the MOOC grader.

The old feedback templates got variables from the runactions module, which used the old sandbox action configurations from the exercise YAML file. Since the sandbox actions are no longer configured in that way, the feedback templates can not really use exactly the same structure with container graders. That only means that the template should be written a little differently with different variables.

I plan to test this a bit more next week with my own use cases, so no need to merge right away.


raphendyr:
Thanks. Looked very simple change.

PR #21, opened 2018-05-31
Add lang argument support and json form translations

teemulehtinen:
Extend the request parameter support for language versions from exercise view to model views and template views.

### DIFF
--- a/access/views.py
+++ b/access/views.py
@@ -65,35 +65,27 @@ def exercise(request, course_key, exercise_key):
     '''
     post_url = request.GET.get('post_url', None)
     lang = request.GET.get('lang', None)
-
-    # Fetch the corresponding exercise entry from the config.
-    (course, exercise) = config.exercise_entry(course_key, exercise_key, lang=lang)
-    if course is None or exercise is None:
-        raise Http404()
-
-    # Exercise language.
-    if not lang:
-        lang = course.get('lang', DEFAULT_LANG)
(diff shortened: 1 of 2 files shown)


access/views.py
```diff
@@ -260,6 +254,17 @@ def generated_exercise_file(request, course_key, exercise_key, exercise_instance
                     raise Http404()
     raise Http404()
 
+
+def _get_course_exercise_lang(course_key, exercise_key, lang_code):
+    (course, exercise) = config.exercise_entry(course_key, exercise_key, lang=lang_code)
+    if course is None or exercise is None:
+        raise Http404()
+    if not lang_code:
+        lang_code = course.get('lang', DEFAULT_LANG)
```
raphendyr: Can we someway get this from the exercise? In other words, do we know the language a specific exercise is done with? In case where different language would be totally different exercise (for some unknown reason). Or does this actually even matter in case of single language for a exercise?

raphendyr:
Looks good.

Can you for the record paste example `exercise_info` here. I can then also check that the syntax looks good too. Based on code, it should, but so I can validate.

teemulehtinen:
Following is an exercise-info json object produced from the o1 course material:
[exercise-info.json.txt](https://github.com/Aalto-LeTech/mooc-grader/files/2079284/exercise-info.json.txt)

raphendyr:
Perfect!

PR #23, opened 2018-06-08
Improve the pregenerate_exercises command

Mankro:
This small improvement is especially useful for the run-mooc-grader container.

Do not raise an exception if the given course has no personalized exercises;
the command can then be run with all courses without crashing.

Add an option to only generate new instances if none exist yet.
This allows the run-mooc-grader container to use the command and
still preserve any previously generated and stored instances.

### DIFF
--- a/access/management/commands/pregenerate_exercises.py
+++ b/access/management/commands/pregenerate_exercises.py
@@ -3,9 +3,12 @@ import glob
 from django.conf import settings
 from django.core.management.base import BaseCommand, CommandError
 from access.views import config
-from util.personalized import delete_pregenerated_exercise_instances, \
-    prepare_pregenerated_exercises_directory, generate_exercise_instances, \
-    pregenerated_exercise_instances
+from util.personalized import (
+    delete_pregenerated_exercise_instances,
+    generate_exercise_instances,
+    pregenerated_exercise_instances,
+    prepare_pregenerated_exercises_directory,
+)
 
 class Command(BaseCommand):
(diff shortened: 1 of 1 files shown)


access/management/commands/pregenerate_exercises.py
```diff
@@ -42,14 +46,19 @@ def handle(self, *args, **options):
             # take only personalized exercises
             exercises = list(filter(lambda ex: "personalized" in ex and ex["personalized"], exercises))
             if not exercises:
-                raise CommandError("The course %s has no personalized exercises" % (course_key))
+                self.stderr.write("The course %s has no personalized exercises so no instances are generated." % (course_key))
```
raphendyr: something like `self.stdout.write(self.style.INFO(...))` could be better. Also, no need to use stderr

access/management/commands/pregenerate_exercises.py
```diff
@@ -4,7 +4,8 @@
 from django.core.management.base import BaseCommand, CommandError
 from access.views import config
 from util.personalized import delete_pregenerated_exercise_instances, \
-    prepare_pregenerated_exercises_directory, generate_exercise_instances
+    prepare_pregenerated_exercises_directory, generate_exercise_instances, \
+    pregenerated_exercise_instances
```
raphendyr: Import could be reformatted to

```python
from foobar import (
  baz1,
  baz2,
  baz3,
)
```
comma in every line and words in sorted order.

PR #26, opened 2018-07-16
use course key as cwd for generator cmd by default

piehei:
This change makes it possible for pregenerating personalized exercises both in local environment powered by course-templates docker/docker-compose setup *and* in production environment.

When the course is developed locally, it is located in mooc-grader file system in courses/default. When the course is deployed to production mooc-grader server, the location is almost certainly different, for instance courses/wsd-2018. Because the paths are different the "cwd" parameter of "generator" in exercise.yml doesn't help, it would need to be rewritten depending on the environment.

My proposal changes the behaviour of **_manage.py pregenerate_exercises <course_key>_** so that the "cmd" of "generator" will be executed from courses/<course_key>, not from the mooc-grader's root path where manage.py resides.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ scripts/chroot_execvp
 
 .DS_Store
 *~
+venv
--- a/courses/README.md
+++ b/courses/README.md
@@ -98,8 +98,12 @@ Durations are given in (int)(unit), where units are y, m, d, h or w.
 		   of the generated file
 		* `allow_download`: if true, the generated file can be downloaded from the web
 	* `generator`: (required if personalized) settings for the generator program that
-		creates one new instance of the exercise. At least `cmd` must be set.
+		creates one new instance of the exercise. At least `cmd` must be set. The generator
+		cmd will be run from course_key dir (course_key is the cwd).
 		* `cmd`: command line as an ARRAY that is used to run the generator.
+			Eg. ["generator_script.sh"] will run generator_script.sh from course_key dir
+			and ["python3", "script_dir/generator.py"] will run generator.py from
+			course_key/script_dir but keep course_key as cwd.
 			Mooc-grader appends the instance directory path to the argument list and
 			the generator is expected to write files into the directory. The file names
 			should be listed under `generated_files` setting so that mooc-grader is aware
@@ -107,8 +111,9 @@ Durations are given in (int)(unit), where units are y, m, d, h or w.
 			`python manage.py pregenerate_exercises course_key <exercise_key>`
 			(`--help` option prints all possible arguments).
 		* `cwd`: if set, this sets the current working directory for the generator
-			program. Start the path from the course directory (course key as the
-			first directory).
+			program. Since the default cwd is course_key, this applies to directories
+			in course_key. Eg. cwd: "script_dir" will change the cwd to course_key/script_dir
+			and only after that run cmd.
 	* `max_submissions_before_regeneration`: (optional, only usable in personalized exercises)
 		defines how many times the student may submit before the personalized exercise is
 		regenerated (the exercise instance is changed to another one). If unset,
--- a/util/personalized.py
+++ b/util/personalized.py
@@ -206,12 +206,13 @@ def generate_one_exercise_instance(course, exercise, dir_path):
         raise access.config.ConfigError(
                 'Missing "generator" and/or "cmd" under "generator" in the exercise configuration %s/%s' %
                 (course["key"], exercise["key"]))
-    
-    cwd = None
+
+    # default cwd is the course directory
+    cwd = os.path.join(access.config.DIR, course["key"])
     if "cwd" in exercise["generator"]:
-        # cwd path in the exercise config should start from the course directory
-        cwd = os.path.join(access.config.DIR, exercise["generator"]["cwd"])
-    
+        # if cwd is set, it should reside inside course directory
+        cwd = os.path.join(cwd, exercise["generator"]["cwd"])
+
     command = exercise["generator"]["cmd"][:] # copy the command list from config before appending
     command.append(dir_path)
     return invoke(command, cwd)


raphendyr:
That path is read only inside local test container, so it won't help there. This should be defined inside django settings, so location can be easily defined to something with write access.

p.s. `manage.py pregenerate_exercises` is "deprecated" and will be removed with roman/godfather/etc. in a year at latest.

Mankro:
Using the course directory sounds like a good default value for "cwd". The "cwd" field should also recognize paths starting with `./` n which case the dot is automatically replaced with the course key (maybe add that to this PR?). Then, changing the course key does not break the configuration.

@raphendyr I do not understand your comment: "That path is read only inside local test container, so it won't help there." How is the course path a problem?

piehei:
I'm not sure I'm following.

We have a server that runs mooc-grader. In the filesystem of that server we have a directory /srv/grader/mooc-grader/courses which has one directory for each course that uses the mooc-grader server.

So inside courses we have eg. wsd-2018, prog1-2018, dbms-2018 etc. When those courses are developed locally with the docker-compose setup, the paths are always courses/default. So the automatic generation of personalized exercises doesn't work the same as the course directory changes from local development to production deployment.

Or am I missing something (might be of course!)?

Mankro:
The purpose of the "cwd" for the generator field is to set the current working directory of the generator process to such a value that the generator program may easily refer to the file paths it needs. Cwd may be set to the course directory or a subdirectory under it. Using the course directory is a good default value and adding the dot format `./` would make it easy to specify a subdirectory without hardcoding the course key in the YAML file.

raphendyr:
Path `/srv/grader/courses/default` is mounted in read-only mode, thus it doesn't have COW layer on top of it (mount, not a docker layer).

ref: https://github.com/apluslms/course-templates/blob/master/docker-compose.yml#L10

Alternatively, if development code is used, then whole `/srv/grader/mooc-grader` and thus `/srv/grader/mooc-grader/courses/default` is in read-only mode. So none of this paths is writable inside the container.

Mankro:
@raphendyr But the generator is not expected to write to the course directory. It should write to the "exercises-meta" directory that is outside the "courses" directory.

piehei:
@raphendyr I think we're talking about different things. See @Mankro 's comment.

I have no problem with writing the personalized files with my generator script. I have only a problem with discovering the script (calling it deterministically, if you will).

raphendyr:
in that case, notice this line https://github.com/Aalto-LeTech/mooc-grader/blob/master/grader/settings.py#L170

piehei:
That is NOT the problem, either :D There is *NO PROBLEM* with the location of the files the generator ### **_produces_**.

raphendyr:
I see. Yeah. I'm stupid. Well, this is probably good enough when I read what the rest of the code does. access.DIR should be in the settings, but isn't yet :(

Carry on, @Mankro had good point about `./` stuff.

Mankro:
The cwd field in the exercise config.yaml has nothing to do with the "exercises-meta" directory. The generator program does not write to the cwd path. It receives the writing path as an argument, while cwd is kind of the reading path so that the generator can easily read files from the course repository without complicated file paths.

piehei:
I think the optimal solution would be:

No matter what you write in cmd of generator in exercise.yml, the cmd will be executed with cwd being the course key directory.

The cwd option only complicates this. Using the "./" convention also complicates it because:
cmd: ["python3", "my_generator_script.py"]
cannot be prepended with "./".

I think there should be no reason for anyone ever to have the generator NOT reside inside the course key directory. Thus the cwd should always be that directory.

What do you think?

raphendyr:
Sounds good.

Mankro:
The default value for "cwd" can be the course directory when it is not specified in config.yaml. If one wants to specify a subdirectory of the course directory, then the dot shorthard `./` would be very useful. I don't see how the dot complicates anything: it is only a shorthand and the code replaces the dot with the course key before before the generator subprocess is started. It only takes one more if statement to support the dot: like `if cwd.startswith('./'): cwd = course['key'] + cwd[1:]` (cwd is the string value here)

piehei:
I think it complicates things because the more knobs the system gives to the user to tune, the more overall knobs there exist and the more overwhelming the documentation page seems to the reader. Since it is completely doable in any scripting language to change cwd to some subdirectory when the script is called, my feeling is that leaving the cwd option and ./ notation completely out would make configuration of the yaml files a tad more straightforward and less burdensome. I see this feature only as a nice-to-have but not necessarily needed.

Implementation-wise you're completely right @Mankro , it isn't too many lines of code!

***If*** the cwd option needs to stay my proposal would be:

1. default mode, no cwd specified
- no cwd specified --> cwd is course_key
- cmd is of form ["script.sh"] --> runs script.sh from the course_key dir
- cmd is of form ["some_dir/script.sh"] --> runs scrip.sh from course_key/some_dir while keeping course_key as cwd

2. cwd is specified
- if cwd is specified --> cwd is what the cwd param says *inside* the course_key
- cwd is of form "dir1/dir2" --> cwd is now course_key/dir1/dir2
- cmd if of form ["script.sh"] --> changes cwd to course_key/dir1/dir2 & runs script.sh
- cmd is of form ["dir3/script.sh"] --> chnages cwd to course_key/dir1/dir2 & runs dir3/script.sh -> dir3 should obviously reside in dir2 but during the execution the cwd is dir2

I think with this one can achieve anything. Please note I've completely left out the ./ stuff since if everything is relative to course_key, there shouldn't be any need for ./. If you're in course_key, you wouldn't say something like cd ./my_dir, you would say cd my_dir. If the default cwd didn't change to be course_key then ./ might propably offer something useful.

What do you think?
:)

raphendyr:
Do that. looks good!

Mankro:
I support this too. Remember to update the `courses/README.md` file too.

raphendyr:
And if it get's too complicated, then let it be. As said, it will be removed at some point :)

piehei:
Might be ready now. Check out the readme changes!

Mankro:
Looks good to me.

piehei:
@raphendyr hop

Issue #28, opened 2018-07-23
Store language in the form. so when it's reposted, the grading is done with the same language


raphendyr:
When we create form for GET request, we should record lang parameter in the form, so when that form is reposted, then we use the same language.

raphendyr:
Fixed by

PR #29, opened 2018-08-10
Add lang field to GradedForm

jrp6:
and add a special case for hidden fields in graded_form.html.
Get language from POST if available.

This handles the language mismatch bug in every case that applies to O1,
except for the piazza exercise. Because whether A+ loads the previous submission
is decided by the course material and not grader, it is almost impossible to
fix the bug in general.

### DIFF
--- a/access/templates/access/graded_form.html
+++ b/access/templates/access/graded_form.html
@@ -8,6 +8,8 @@
 
 		{% if field.field.type == "static" %}
 		{{ field.field.more|safe }}
+		{% elif field.field.type == "hidden" %}
+			{{ field }}
 		{% else %}
 
 		{% if not field.field.row_label or field.field.open_table %}
--- a/access/types/forms.py
+++ b/access/types/forms.py
@@ -142,6 +142,13 @@ class GradedForm(forms.Form):
 
             g += 1
 
+        if 'lang' in self.exercise:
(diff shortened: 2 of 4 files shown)


access/templates/access/graded_form.html
```diff
@@ -8,6 +8,8 @@
 
 		{% if field.field.type == "static" %}
 		{{ field.field.more|safe }}
+    {% elif field.field.type == "hidden" %}
```
raphendyr: It seems here is different indentation characters than in other places :/

access/types/forms.py
```diff
@@ -142,6 +142,13 @@ def __init__(self, *args, **kwargs):
 
             g += 1
 
+        if 'lang' in self.exercise:
+            self.add_field(i, {
+                'type': 'hidden',
+                'initial': self.exercise['lang'],
+                'key': 'lang',
```
raphendyr: This probably should be something like `__lang` or `__grader_lang`, so there won't be accidental name collisions.

access/templates/access/graded_form.html
jrp6: The editor didn't recognize the file being indented with tabs not spaces, my bad.

raphendyr:
Fixes #28

PR #35, opened 2019-02-08
submission acceptGitAddress cookie 'aplusgit' did not work in newer browsers

jykesaviaGNH:
Newer browsers (chromium 71, firefox 60) did not fire
DOMContentLoaded event (doc is already complete when <script> is running).

Setup now looks at 'document.readyState' and will register
the click-callback immediately if document is no longer in 'loading' state.
info from:
https://stackoverflow.com/questions/43233115/chrome-content-scripts-arent-working-domcontentloaded-listener-does-not-execut

Tested: jslint, chromium latest, firefox longterm support

### DIFF
--- a/access/templates/access/accept_git_form.html
+++ b/access/templates/access/accept_git_form.html
@@ -26,12 +26,12 @@ var aplusSetGitCookie = function($) { // supply jQuery as parameter
 	}
 
 	$("#git_submit").on("click", function() {
-		var id = $("#git_id").val();
-		if (id.trim().length > 0) {
+		var id_val = $("#git_id").val();
+		if (id_val.trim().length > 0) {
 			var expire = new Date();
 			expire.setFullYear(expire.getFullYear() + 1);
 			var path = document.location.href.split("/");
-			document.cookie = "aplusgit=" + encodeURIComponent(id)
+			document.cookie = "aplusgit=" + encodeURIComponent(id_val)
 				+ ";path=/" + path[3] + "/" + path[4] + "/"
 				+ ";expires=" + expire.toGMTString();
 		}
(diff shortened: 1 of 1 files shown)


raphendyr:
Two things. First, code could hook to `aplus:exercise-loaded` event, like this single post hook does https://github.com/Aalto-LeTech/a-plus/blob/master/exercise/static/exercise/chapter.js#L836

Second, we probably should use local storage instead of cookies. The information is not required at the server. Example is in how LTI page does it https://github.com/Aalto-LeTech/a-plus/blob/master/external_services/static/js/external_launcher.js . The local storage value should probably be visible in profile page for removal, similar to how LTI values are. Though, this probably should be implemented as generic and not specific to this use case.

Hmm.. Actually, the `exercise-ready`might be better here, but both should work as well. That even API is not yet documented or well designed for A+. I hope we get person to do that during the summer.

raphendyr:
Hmm.. At bare minimum, the file should be wrapped to a private namespace. So, instead of merging this bug fix (should be done, but introduces another global function to A+ javascript global namespace), I request a redesign of the code. Following thins should be implemented:

* Wrap code in a private namespace
* Remove `#git_id`, as we can't use id parameters, when there are multiple of this in a same page.
* Listen for aplus event, so we will get a exercise element and can bind to the form with a exercise.
* Use unique random string to identifie a form instance, so the JS will target only a single form. In chapter with multiple of these forns, there will be multiple instances of the listeners running, and they will receive events from all of the exercises. Thus, the code for this exercise, must notice the event for this exrcise. This is a lot of complication until A+ supports javascript plugins, so we get only single instance of this code running.
* Remove cookie and store data in a local storage. A+ should be updated to show the local storage value nicely (currently only shoes values for lti redirect pages).

That should be pretty much it.

jykesaviaGNH:
Don't have time and skill to fix this.

PR #36, opened 2019-05-28
change cron.sh cron_pull_build.sh

JohnDing1995:
This is only for compare cron.sh cron_pull_build.sh


JohnDing1995:
Now allowing user to use DOCKER_HOST_PATH environment variable as host path(this path not includes course key)

Mankro:
This old pull request is no longer relevant. The gitmanager app has been removed from the mooc-grader in v1.12. Git Manager is a separate service now. The old Shepherd (godfather) and Roman projects were discontinued.

PR #38, opened 2019-07-23
Grader API

QianqianQ:
Deploy API design with the help of apluslms_file_transfer package


grader/ajax_key.py
```diff
@@ -0,0 +1,6 @@
+"""
```
raphendyr: Hm.. This and `secret_key.py` should probably be in .gitignore :thinking: . Neither of them should be in the git.

grader/settings.py
```diff
@@ -9,7 +9,7 @@
 
 # Base options, commonly overridden in local_settings.py
 ##########################################################################
-DEBUG = False
+DEBUG = True
```
raphendyr: You can set `GRADER_DEBUG=true` environment variable for the container. Any case, this should stay as false.

grader/settings.py
```diff
@@ -250,3 +252,24 @@
 
 # update template loaders for production
 use_cache_template_loader_in_production(__name__)
+
+
+###############################################################################
+# JWT settings
+JWT_PUBLIC_KEY = """
```
raphendyr: This is also something, that should be passed via the environment.

static/CS-A4444_test5
```diff
@@ -0,0 +1 @@
+/u/71/qinq1/data/Desktop/mooc-grader/courses/CS-A4444_test5/_build/html
```
raphendyr: These files shouldn't ne in the git.

QianqianQ:
To improve: post the sent file by chunks

requirements.txt
```diff
@@ -7,3 +7,5 @@ lxml==3.6.0
 PyYAML==3.11
 docutils==0.12
 python-magic==0.4.11
+PyJWT==1.7.1
```
raphendyr: use `PyJWT[crypto] ~= 1.7.1` here and drop `cryptography`. The `[crypto]` includes all needed requirements.

api/urls.py
```diff
@@ -0,0 +1,6 @@
+from django.conf.urls import url
+from api import views
+
+urlpatterns = [
+    url(r'^$', views.course_handler, name='course_handler'),
```
raphendyr: Can you prefix these with `v1`. Also, it's a good idea to have a sub path for this simple file manipulation (if we require more, we can add different paths easily). Thus, the full path would be `/api/v1/files/` for example. Last part could be `files` or `upload` for example.

api/utils.py
```diff
@@ -0,0 +1,89 @@
...
+from django.core.exceptions import (
+    ImproperlyConfigured,
+    PermissionDenied,
+)
+
+logger = logging.getLogger(__name__)
+
+# ----------------------------------------------------------------------------------------------------------------------
+# JWT Authentication
+# https://github.com/Aalto-LeTech/django-lti-login/blob/jwt-api/django_lti_login/api_views.py
```
raphendyr: This time, don't include this reference, as I will remove that branch once the feature is merged in lti login. Normally, it's good style to reference the source :)

raphendyr: I kind of would like to include the course path in the url. e.g. `/api/v1/def_course/upload`, which would make the jwt_auth part a bit nicer :)

api/utils.py
```diff
@@ -0,0 +1,89 @@
...
+                       key=key,
+                       algorithms=settings.JWT_ALGORITHM,
+                       **options)
+    return None
+
+
+jwt_decode = prepare_decoder()
+
+
+def jwt_auth(request):
```
raphendyr: You can refactor this to a decorator, and then it can be used like django's login_required (https://github.com/django/django/blob/master/django/contrib/auth/decorators.py#L38)

api/utils.py
```diff
@@ -0,0 +1,89 @@
...
+                     request.META.get('HTTP_AUTHORIZATION', ''))
+        raise PermissionDenied("Invalid authorization header")
+
+    # decode jwt token
+    try:
+        return jwt_decode(token)
+    except jwt.InvalidTokenError as exc:
+        logger.debug("JWT auth failed: %s", exc)
+        raise PermissionDenied(str(exc))
+
```
raphendyr: If the course directory (name) is also in the api url, then it should be validated inside the jwt_auth-

api/views.py
```diff
@@ -0,0 +1,185 @@
...
+
+    if request.method == 'POST' or request.method == 'PUT':
+        # POST: upload a new course  PUT: upload a new course / update an existing course
+
+        if request.method == 'POST':
+            # The course already exists
+            if os.path.exists(course_directory):
+                return JsonResponse({'error': "The course has been uploaded"}, status=400)
+
+        # parse data
```
raphendyr: This parsing logic is only required for `PUT`. Django handles this for `POST` and the data should be available in the `request.POST` and `request.FILES`

raphendyr: I think for this api the `POST` should be enough. Use it for new and existing course.

api/views.py
```diff
@@ -0,0 +1,185 @@
...
+        else:
+            logger.warning(content_type)
+            return HttpResponseBadRequest("Unsupported content-type")
+
+        # Upload/ Update the course directory in mooc-grader
+        try:
+            temp_course_dir = os.path.join(course_path, 'temp_' + auth['sub'].strip())
+
+            if content_type == 'application/octet-stream':
+
```
raphendyr: I think this part can be refactored with support functions and/or classes.

grader/settings.py
```diff
@@ -22,16 +22,18 @@
 
 INSTALLED_APPS = (
     # 'django.contrib.admin',
-    # 'django.contrib.auth',
-    # 'django.contrib.contenttypes',
+    'django.contrib.auth',
```
raphendyr: What requires this?

QianqianQ: Previously I also used django rest framework to build the api and it needs these. I removed the api using drf later but forgot to modify here. Now I comment them.

QianqianQ: > What requires this?

Previously I also used django rest framework to build the api and it needs these. I removed the api using drf later but forgot to modify here. Now I comment them.

QianqianQ:
Add comments, Reformat code and Testing cases

QianqianQ:
To do: Testing

PR #42, opened 2019-09-12
Modify the grading of  the checkbox forms.

alalav1:
Add the partial_points setting, a possibility to have a model answer where all
options are wrong and a possibility make neutral options.  Feedback always gives max points in questionnaires in general.

### DIFF
--- a/access/types/forms.py
+++ b/access/types/forms.py
@@ -318,6 +318,9 @@ class GradedForm(forms.Form):
         if hint and not hint in hints:
             hints.append(hint)
 
+    def is_enrollment_exercise(self):
+        return self.exercise.get('status', '') in ('enrollment', 'enrollment_ext')
+
     def bind_initial(self):
         '''
         Binds form using initial values.
@@ -353,6 +356,7 @@ class GradedForm(forms.Form):
                     if gname not in error_groups:
                         error_groups.append(gname)
             g += 1
+
         return (points, error_groups, error_fields)
 
     def compare_values(self, method, val, cmp):
@@ -450,7 +454,15 @@ class GradedForm(forms.Form):
                 else:
                     ok, hints, method = self.grade_checkbox(
                         self.row_options(configuration, row), value, hints)
-                if ok:
+                if self.exercise.get("feedback") or self.is_enrollment_exercise():
+                    points += row.get("points", 0)
+                    max_points += row.get("points", 0)
+                elif self.row_options(configuration, row).get("partial_points"):
+                    points += max(int(row.get("points", 0) * ok), 0)
+                    max_points += row.get("points", 0)
+                    if not points == max_points:
+                        all_ok = False
+                elif ok:
                     points += row.get("points", 0)
                     max_points += row.get("points", 0)
                 else:
@@ -474,6 +486,23 @@ class GradedForm(forms.Form):
             ok, hints, method = True, [], 'string'
         else:
             raise ConfigError("Unknown field type for grading: %s" % (t))
+
+        points = configuration.get('points', 0)
+        if self.exercise.get("feedback", False) or self.is_enrollment_exercise():
+            # Feedback questionnaires grant full points unless they are rejected.
+            # The partial_points is not meant to be used in feedback questionnaires.
+            # However, if they both exist, the grader should not crash.
+            ok = True
+            earned_points = points
+        elif configuration.get('partial_points'):
+            # grade_checkbox returns ok as float instead of a boolean, if
+            # 'partial points' is set
+            earned_points = max(int(points * ok), 0)
+            ok = (earned_points == points)
+        elif ok:
+            earned_points = points
+        else:
+            earned_points = 0
 
         # Apply new feedback definitions.
         for fb in configuration.get("feedback", []):
@@ -499,12 +528,11 @@ class GradedForm(forms.Form):
             if add:
                 hints.append(new_hint)
 
-        points = configuration.get('points', 0)
         if name in self.fields:
-            self.fields[name].grade_points = points if ok else 0
+            self.fields[name].grade_points = earned_points
             self.fields[name].max_points = points
             self.fields[name].hints = hints
-        return i + 1, ok, points if ok else 0
+        return i + 1, ok, earned_points
 
     def row_options(self, configuration, row):
         hint = row.get('hint', '')
@@ -522,27 +550,50 @@ class GradedForm(forms.Form):
     def grade_checkbox(self, configuration, value, hints=None):
         hints = hints or []
         correct_count = 0
-
-        # All correct required if any configured
+        wrong_answers = 0
+        non_neutral_count = 0
+        # If partial points is not set, all options must be answered correctly
+        # in order to gain points.
+        # If no correct answers are set in configuration, points are granted to
+        # an empty answer only.
         correct = True
         i = 0
         for opt in configuration.get("options", []):
             name = self.option_name(i, opt)
-            if opt.get("correct", False):
+            correct_answer = opt.get("correct", False)
+            # correct_answer may be boolean or string "neutral"
+            if correct_answer is True:
                 correct_count += 1
+                non_neutral_count += 1
                 if name not in value:
+                    wrong_answers += 1
                     correct = False
                     self.append_hint(hints, opt)
-            elif not value is None and name in value:
-                correct = False
-                self.append_hint(hints, opt)
+            elif correct_answer is False:
+                non_neutral_count += 1
+                if value is not None and name in value:
+                    correct = False
+                    wrong_answers += 1
+                    self.append_hint(hints, opt)
+            elif correct_answer == "neutral":
+                if value is not None and name in value:
+                    self.append_hint(hints, opt)
             i += 1
+
+        # If partial_points are set, the variable 'correct' becomes a float
+        # less than one instead of a boolean. It will be used to multiply
+        # the max points.
+        if configuration.get("partial_points"):
+            if wrong_answers == 0 or non_neutral_count == 0:
+                correct = 1
+            else:
+                correct = (non_neutral_count / 2.0 - wrong_answers) / (non_neutral_count / 2.0)
 
         # Add note of multiple correct answers.
         if correct_count > 1 and len(value) == 1:
             hints.append(_("Multiple choices are selectable."))
 
-        return correct_count == 0 or correct, hints, 'array'
+        return correct, hints, 'array'
 
     def grade_radio(self, configuration, value, hints=None):
         hints = hints or []
--- a/access/templates/access/graded_form.html
+++ b/access/templates/access/graded_form.html
@@ -28,7 +28,9 @@
 			{% if not field.field.group_errors %}
 			{% if result.accepted %}
 				{% if not exercise.feedback or field.field.points > 0 %}
-				<span class="badge badge-{% if field.field.name in result.error_fields %}danger{% else %}success{% endif %}">{{ field.field.grade_points }} / {{ field.field.max_points }}</span>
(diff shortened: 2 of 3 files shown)


Mankro:
@alalav1 please rebase against the upstream master

access/types/forms.py
```diff
@@ -322,12 +322,18 @@ def grade(self):
...
+                    self.fields[self.field_name(prev, field)].grade_points = field.get('points', 0)
+                elif not ok:
                     error_fields.append(self.field_name(prev, field))
                     gname = self.group_name(g)
                     if gname not in error_groups:
                         error_groups.append(gname)
             g += 1
+
+        if self.exercise["feedback"]:
+            points = self.exercise["max_points"]
```
Mankro: Change `self.exercise["max_points"]` into `self.exercise.get("max_points", 0)` just in case max_points are not set.

access/types/forms.py
```diff
@@ -452,6 +458,12 @@ def grade_field(self, i, configuration):
         else:
             raise ConfigError("Unknown field type for grading: %s" % (t))
 
+        points = configuration.get('points', 0)
+        points_got = points if ok else 0
+        if configuration.get('partial_points'):
+            points_got = max(int(points*ok),0)
```
Mankro: Add spaces around the multiplication operator ` * `.

access/types/forms.py
```diff
@@ -452,6 +458,12 @@ def grade_field(self, i, configuration):
         else:
             raise ConfigError("Unknown field type for grading: %s" % (t))
 
+        points = configuration.get('points', 0)
+        points_got = points if ok else 0
```
Mankro: `points_got` could be renamed `earned_points`

access/types/forms.py
Mankro: It is confusing that `ok` sounds like a boolean variable, but you use it like a number in a multiplication. What is the type of `ok`?

access/types/forms.py
```diff
@@ -499,27 +510,40 @@ def row_options(self, configuration, row):
...
                 if name not in value:
+                    wrong_answers += 1
                     correct = False
                     self.append_hint(hints, opt)
-            elif not value is None and name in value:
-                correct = False
-                self.append_hint(hints, opt)
+            elif correct_answer is False:
+                non_neutral_count += 1
+                if not value is None and name in value:
```
Mankro: I prefer `value is not None` over `not value is None`.

access/types/forms.py
```diff
@@ -499,27 +510,40 @@ def row_options(self, configuration, row):
...
+                    correct = False
+                    wrong_answers += 1
+                    self.append_hint(hints, opt)
             i += 1
 
+        #if partial_points are set, correct is scaled to 0-1 float, to be factored to max points
```
Mankro: The formatting of these comments needs to be fixed.

access/types/forms.py
```diff
@@ -499,27 +510,40 @@ def row_options(self, configuration, row):
     def grade_checkbox(self, configuration, value, hints=None):
         hints = hints or []
         correct_count = 0
-
-        # All correct required if any configured
+        wrong_answers = 0
+        non_neutral_count = 0
+        # All correct required if any configured and partial-points not set
```
Mankro: `# All correct required if any configured and partial-points not set`
Is the comment correct anymore? Does the new implementation fail all submissions if the config has no correct choices set (in normal non-feedback questionnaires)? I think we decided to make it so that no submissions are accepted if the teacher forgets to mark up the correct choices. Previously, the grader awarded any submission full points in that case.

access/types/forms.py
```diff
@@ -499,27 +510,40 @@ def row_options(self, configuration, row):
...
+                if not value is None and name in value:
+                    correct = False
+                    wrong_answers += 1
+                    self.append_hint(hints, opt)
             i += 1
 
+        #if partial_points are set, correct is scaled to 0-1 float, to be factored to max points
+        #The zero points limit is in the half of the correct answers
+        if configuration.get("partial_points"):
+            correct = float(non_neutral_count / 2 - wrong_answers) / (non_neutral_count / 2)
```
Mankro: If there is only one choice (checkbox), is this going to divide by zero?

access/types/forms.py
```diff
@@ -452,6 +458,12 @@ def grade_field(self, i, configuration):
         else:
```
Mankro: GitHub does not let me comment on lines that were not modified. Anyway, in this `grade_field` method, the if statement with "table-radio" and "table-checkbox" uses the `self.grade_checkbox` method too. Its return value was changed, so the method call should be updated everywhere where the method is used.

access/types/forms.py
```diff
@@ -322,12 +322,18 @@ def grade(self):
                 prev = i
                 i, ok, p = self.grade_field(i, field)
                 points += p
-                if not ok:
+                if self.exercise["feedback"]:
+                    self.fields[self.field_name(prev, field)].grade_points = field.get('points', 0)
```
Mankro: I think that `self.fields[...].grade_points` should be set in the `grade_field` method since it is normally done there. Why should it be done here for feedback exercises separately?

Mankro:
We have to check how feedback questionnaires behave now that their questions are graded as incorrect, but the submission receives max points for the questionnaire anyway. We recently changed how the points are rendered in the exercise HTML, so we need to check if the points are rendered in a bad way now.

@alalav1 There are some comments that require actions, but overall, this looks nice.

access/types/forms.py
alalav1: Yes. There is no sense to use "partial_points" with just one non_neutral. This situation can occur in randomized checkbox questions so this should still be fixed.

alalav1:
> We have to check how feedback questionnaires behave now that their questions are graded as incorrect, but the submission receives max points for the questionnaire anyway. We recently changed how the points are rendered in the exercise HTML, so we need to check if the points are rendered in a bad way now.
> 
> @alalav1 There are some comments that require actions, but overall, this looks nice.

The grade_field option doesn't yet affect any html or templates or anything else front-end related, it only returns the values needed for that. The conditional statement the feedback exercises give all the points regardless of anything Before marking the incorrect fields or grops. The `ok` return value is never actually checked for the feedback questions, thus the html should work as expected.

access/types/forms.py
alalav1: Indeed if the ok value was changed in the grade_field, this if else statement could be more clear and it would do the grading where it should be done.

alalav1:
> > We have to check how feedback questionnaires behave now that their questions are graded as incorrect, but the submission receives max points for the questionnaire anyway. We recently changed how the points are rendered in the exercise HTML, so we need to check if the points are rendered in a bad way now.
> > @alalav1 There are some comments that require actions, but overall, this looks nice.
> 
> The grade_field option doesn't yet affect any html or templates or anything else front-end related, it only returns the values needed for that. In the conditional statement the feedback exercises give all the points before marking the incorrect fields or grops. The `ok` return value is never actually checked for the feedback questions, thus the html should work as expected. And I actually tested that, too.

However the if statement will be probably reverted and the `ok` value is modified in the feedback questionnaires. In that case the html should treat all answers as correct as well.

alalav1:
Apparently I was wrong in my previous comment, grade field actually modifies some field attributes itself, and can actually modify html output. I am going to change the logic and move the feedback handling to `
`grade_field` method

access/types/forms.py
alalav1: Now it will award empty checkbox with full points. If there are correct choices, every one of the must be chosen to get any points, if partial-points is not set

access/types/forms.py
alalav1: Rare occasion but still possible, especially with randomized checkbox. Can be avoided with converting 2 to float 2.0.

access/types/forms.py
```diff
@@ -450,7 +454,15 @@ def grade_field(self, i, configuration):
                 else:
                     ok, hints, method = self.grade_checkbox(
                         self.row_options(configuration, row), value, hints)
-                if ok:
+                if self.exercise.get('feedback'):
+                    points += row.get("points", 0)
+                    max_points += row.get("points", 0)
+                elif "partial_points" in self.row_options(configuration, row):
```
Mankro: `partial_points` may be False in the config.yaml so it is not enough to check whether the key exists. You should use the `get` method of dictionaries.

access/types/forms.py
```diff
@@ -353,6 +353,10 @@ def grade(self):
                     if gname not in error_groups:
                         error_groups.append(gname)
             g += 1
+
+        if self.exercise["feedback"]:
+            points = self.exercise.get("max_points")
```
Mankro: The default in `get` is None if it is not specified, so you should do `self.exercise.get("max_points", 0)`.

access/types/forms.py
Mankro: Right, if no correct choices are set, then the student gets points by submitting an empty solution. At any rate, the comment in the code is hard to understand and it should be written more clearly.

access/types/forms.py
Mankro: Now that you use floats, if the student answers correctly and `wrong_answers` is zero, is `correct` really guaranteed to equal `1` since floating point arithmetic is inherently inaccurate?

access/types/forms.py
```diff
@@ -450,7 +454,15 @@ def grade_field(self, i, configuration):
                 else:
                     ok, hints, method = self.grade_checkbox(
                         self.row_options(configuration, row), value, hints)
-                if ok:
+                if self.exercise.get('feedback'):
+                    points += row.get("points", 0)
+                    max_points += row.get("points", 0)
+                elif "partial_points" in self.row_options(configuration, row):
+                    points += row.get("points", 0) * ok
```
Mankro: When you multiply with `ok`, shouldn't the result be casted to `int` since it becomes a float otherwise?

access/types/forms.py
```diff
@@ -475,6 +487,19 @@ def grade_field(self, i, configuration):
         else:
             raise ConfigError("Unknown field type for grading: %s" % (t))
 
+        if self.exercise.get("feedback"):
+            # The partial_points is not meant to be used in feedback questionnaires.
+            # However if they both exist the grader should not crash.
+            ok = 1
```
Mankro: If `ok` becomes `1` here (if the exercise is a feedback questionnaire), then does this method return `1` for `ok`? The caller expects the value to be a boolean, so can't you keep it boolean? You can prevent the partial points scaling for feedback exercises in other ways too.

access/types/forms.py
```diff
@@ -522,27 +546,40 @@ def row_options(self, configuration, row):
...
-            if opt.get("correct", False):
+            correct_answer = opt.get("correct", False)
+            # correct_answer may be boolean or string "neutral"
+            if correct_answer is True:
                 correct_count += 1
+                non_neutral_count += 1
                 if name not in value:
+                    wrong_answers += 1
                     correct = False
                     self.append_hint(hints, opt)
```
Mankro: Is it possible to define feedback hints for correct answers (when the student answers correctly)?

Is it possible to define feedback hints to neutral alternatives?

access/types/forms.py
```diff
@@ -522,27 +546,40 @@ def row_options(self, configuration, row):
...
+            elif correct_answer is False:
+                non_neutral_count += 1
+                if value is not None and name in value:
+                    correct = False
+                    wrong_answers += 1
+                    self.append_hint(hints, opt)
             i += 1
 
+        # If partial_points are set, correct is scaled to 0-1 float, to be
+        # factored to max points.
```
Mankro: I'm not sure the word "factored" is used correctly in the comment. For example, arent't 2 and 6 factors to 12 because `2 * 6 = 12`? I was thinking that you are actually scaling the points with the float number.

Mankro:
The commit message must be written more clearly.

PR #43, opened 2019-09-13
Validate the submission value types in freetext questions

alalav1:
In freetext-questions non-numeric submissions are rejected,  if the int or float compare-method is used. (Rejected submissions don't affect the amount of submissions left.)
Additionally the error tolerance of the float method is now proportional, instead of a constant 0.02.
New acceptable range is +-1% of the model solutions value, which in general
requires two correct significant digits.

### DIFF
--- a/access/types/forms.py
+++ b/access/types/forms.py
@@ -1,3 +1,4 @@
+import math
 import random
 import re
 import json
@@ -10,6 +11,7 @@ from django.forms.utils import ErrorDict
 from django.forms.widgets import CheckboxSelectMultiple, RadioSelect, Textarea
 from django.utils.safestring import mark_safe
 from django.utils.translation import ugettext as _
+
 from util.files import random_ascii
 from util.templates import template_to_str
 from util import forms as custom_forms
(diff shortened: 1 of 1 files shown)


access/types/stdsync.py
```diff
@@ -143,18 +143,18 @@ def createForm(request, course, exercise, post_url):
     if form.is_valid():
         (points, error_groups, error_fields) = form.grade()
         points = pointsInRange(points, exercise["max_points"])
+        if not 'rejected' in error_groups:
```
Mankro: At first sight, it looks suspicious that `form.is_valid()` would pass for a rejected form. We need to compare this to how checks for required fields are handled, since those can similarly lead to rejected forms.

Mankro: So, we suspect that the required fields are handled by the validation provided by the superclass form (Django forms.Form). You should try to implement custom validation methods on the GradedForm class. They would be called by the `form.is_valid()` method (implemented in the Django superclass). The custom validation method would reject the form if the input can not be parsed as an integer in a text field with the int type/comparator.

https://docs.djangoproject.com/en/2.2/ref/forms/validation/

Mankro: Since GradedForm creates the form fields dynamically based on the exercise configuration YAML, the `form.clean` method might be a suitable place for looping over the fields and checking that integer text fields contain valid integers.  Another alternative is to add validators to the text fields (default field validation runs the validators). The field constructor has a validators parameter.

Mankro: It could be even better to use more descriptive form fields such as IntegerField instead of TextField when we expect integer input. IntegerField also validates the input by default, so we would not need custom validation methods for that.

Mankro:
I fixed some coding style issues and bugs before force-pushing and merging.

PR #44, opened 2019-10-04
Add randomized option for checkbox questions

alalav1:
Also add modification and fixes to pick_randomly questionnaires.

If # is in the end of the question title in the questionnaire, the question
is numbered automatically. This is handy especially when pick_randomly is set
and the order of the questions is not known in the yaml files.

Note: a-plus-rst-tools numerate questions automatically, if question titles are
not set manually. The numerating task could be moved to mooc-grader, so the
pick_randomly works accordingly and all questionnaires are numerated in
the same place.


alalav1:
One thing is yet to be implemented: randomized question can be doesn't reject unrefreshed submissions

Additionally, the automatic numbering and `correct_count` option require minor update to a-plus-rst-tools to be used. (They do not break anything without the update)

alalav1:
> One thing is yet to be implemented: randomized question can be doesn't reject unrefreshed submissions
> 
> Additionally, the automatic numbering and `correct_count` option require minor update to a-plus-rst-tools to be used. (They do not break anything without the update)

If aplus refreshed the form automatically after submitting the answer, the problem would be solved.
However, the problem is likely caused by invalid use of the checksum and nonce parameters so it should be fixed.

Mankro:
@alalav1 It is best that you fix the git conflicts before take a look at this. So, rebase against upstream master and fix the conflicts, please.

Mankro:
When question titles are numbered automatically (particularly for "pick randomly" questionnaires), I think that mooc-grader should check if the title in the config.yaml has the substring `{#}` and replace that with the number. This allows the number to be located anywhere in the string instead of only the end. `{#}` is also a value that is probably never used, whereas somebody might want to use the `#` character literally in a title.

Mankro:
MOOC-grader seems to crash when a client requests the model solution for a pick-randomly questionnaire.

access/types/forms.py
```diff
@@ -67,6 +59,14 @@ def __init__(self, *args, **kwargs):
...
+                # Check that sample is unmodified in a randomized form.
+                if not args[0] is None:
+                    nonce = args[0].get('_nonce', '')
+                    sample = args[0].get('_sample', '')
+                    if self.samples_hash(nonce, sample) != args[0].get('_checksum', ''):
+                        raise PermissionDenied('Invalid checksum')
+                    post_samples = sample.split('/')
+
                 self.randomized = True
                 if not args[0] is None:
```
Mankro: There are two if statements with the same condiftion `if not args[0] is None`. I also prefer `is not` over `not X is ...`

access/types/forms.py
```diff
@@ -211,13 +211,14 @@ def add_table_fields(self, i, config, field_class, widget_class, multiple=False)
 
     def add_field(self, i, config, field_class, widget_class,
             initial=None, correct=None, choices=None, multiple=False,
-            widget_attrs={'class': 'form-control'}):
+            widget_attrs={'class': 'form-control'}, post_data = None):
```
Mankro: `post_data = None` should be `post_data=None`. No space is used with the keyword arguments.

access/types/forms.py
```diff
@@ -233,28 +234,83 @@ def add_field(self, i, config, field_class, widget_class,
         if not choices is None:
             args['choices'] = choices
 
+        name = self.field_name(i, config)
```
Mankro: Could this new long chunk of code be moved to a new method perhaps? Just to make the code more readable and organized.

access/types/stdsync.py
```diff
@@ -131,7 +131,7 @@ def createForm(request, course, exercise, post_url):
         last = max_n > 0 and n >= max_n
 
     form = GradedForm(request.POST or None, request.FILES or None,
-        exercise=exercise, show_correct_once=last)
+        exercise=exercise, show_correct_once=last, request=request)
```
Mankro: The `GradedForm` constructor is also used in other places in the code and the request argument should be added there as well if applicable.

alalav1:
There is a problem with rejecting used forms.

Mankro:
The line `configuration.get('random_sample')` in the `grade_checkbox` method is hacky since the configuration is not supposed to hold data generated in the form (the random sample is generated in the form when the exercise is loaded and then it is used again from the POST data for grading the submission). The configuration dict should only hold the original configuration (yaml) data. However, we can ignore this for the moment and keep this implementation. A better approach could be to use instance variables for the sample, and the sample is actually already in the field instance variables, but the `grade_checkbox` method does not know the name of the field being graded.

Mankro:
I fixed some issues:

* The MOOC-grader gives a proper rejected response to A+ when the random question checksums are invalid. Previously, the HTTP 403 response made A+ show a generic error message and set the submission to the error status even though there is no error in the grader; the submission has forged checksums.
* The nonces were not saved for random checkbox questions, but now they are. This is needed in order to prevent users from reusing previous nonces. Reusing old nonces and checksums would enable users to answer old questions even though the questions are supposed to change randomly.
* The `current_sample` function used to randomly select questions now uses a separate random number generator so that the deterministically set seed does not interfere with other uses of the random module, for example, the nonces.

Mankro:
I fixed more issues that I found:

The question form is now always disabled in the feedback for
random checkbox questions. Previously, it was sometimes enabled
even though the submit button was not rendered.

Random checkbox questions do not generate a new nonce value for
the feedback HTML since the feedback form is disabled and can not
be submitted. The nonce and checksum are not included in the
feedback HTML at all when the form is accepted for grading.

current_sample: the randomly generated sample of question indexes
is not sorted so that the questions can be shown in any random
order. The correct choices in random checkbox questions are not
always listed before the incorrect choices; instead, the order
of the choices is random.

The seed is now based on the exercise and question keys in
addition to the user IDs and the submission ordinal number.
Thus, if the the same exercise is defined multiple times in the
chapter, the random choice of questions is different for each
exercise. Previously, each exercise made the same identical
random sample.

access/types/forms.py
```diff
@@ -605,18 +606,15 @@ def row_options(self, configuration, row):
...
-        # hack: configuration is modified in the method add_field so that
-        # the random_sample can be easily used here
-        randomized_sample = configuration.get('random_sample')
-        is_randomized = bool(randomized_sample)
-        sample = []
-        if is_randomized:
-            sample = [int(x) for x in randomized_sample.split('-')]
+        # Non-random checkbox questions have empty string as random_sample.
+        randomized_sample = self.fields[name].random_sample if name else ''
+        sample = [int(x) for x in randomized_sample.split('-')]
```
Mankro: If an empty string is split, it returns a list with one element, the empty string. Calling `int('')` raises `ValueError`. Did you test this with normal, non-random questions?

Furthermore, note that `bool([""]) == True` and `bool([]) == False`.

PR #45, opened 2019-10-07
Add 'required' html tag to questionnaire fields

alalav1:
Questionnaire fields that are marked as required in yaml files have
'required' html tag in course page


Mankro:
I added the required HTML attribute to the common file upload exercise template too.

PR #46, opened 2019-10-07
Question points are visible before submitting

alalav1:
Add badges to show the question points before submitting in questionnaires.


Mankro:
You added a new language string, so you need to update the file `locale/fi/LC_MESSAGES/django.po` too in order to provide the Finnish translation. Check the Django translation documentation for instructions on how to do that correctly.

Mankro:
@alalav1 I fixed issues in the Finnish translation file and force pushed. However, I noticed that the question points are not always rendered in the way we want. 

If a normal questionnaire has a question with zero points, the feedback renders a gray badge "0 points" instead of the green "0 / 0" badge. Feedback should use the "0 / 0" format since it shows how many points have been earned.

**Feedback questionnaires** should not show gray question point badges with zero points in the exercise description. The badge shall be rendered if the question max points have been set above zero.

Mankro:
I just noticed that enrollment questionnaires typically have zero max points, but this PR shows "0 points" separately for each question. We don't want that since it becomes needless repetition. So, let's not render question points in the exercise description if the whole questionnaire awards zero points, but if some questions award points, then the zero point questions need to show their points in the exercise description too.

We need to check how the points have been rendered in the past in the feedback (after submitting). I guess the question points have usually been hidden if the question is worth zero points.

access/templates/access/graded_form.html
```diff
@@ -27,6 +27,14 @@
 
 			{% if result.accepted and not field.field.group_errors and field.field.max_points > 0 %}
 			<span class="badge badge-{% if field.field.name in result.error_fields %}danger{% else %}success{% endif %}">{{ field.field.grade_points }} / {{ field.field.max_points }}</span>
+			{% else %}
+			<span class="badge">
+				{% blocktrans count points=field.field.points %}
```
Mankro: Is there any reason to use `field.field.points` instead of `field.field.max_points`?

alalav1: The `field.max_points` is defined when grading the field, so it doesn't exist before the first submission.

alalav1:
> I just noticed that enrollment questionnaires typically have zero max points, but this PR shows "0 points" separately for each question. We don't want that since it becomes needless repetition. So, let's not render question points in the exercise description if the whole questionnaire awards zero points, but if some questions award points, then the zero point questions need to show their points in the exercise description too.
> 
> We need to check how the points have been rendered in the past in the feedback (after submitting). I guess the question points have usually been hidden if the question is worth zero points.

alalav1:
> I just noticed that enrollment questionnaires typically have zero max points, but this PR shows "0 points" separately for each question. We don't want that since it becomes needless repetition. So, let's not render question points in the exercise description if the whole questionnaire awards zero points, but if some questions award points, then the zero point questions need to show their points in the exercise description too.
> 
> We need to check how the points have been rendered in the past in the feedback (after submitting). I guess the question points have usually been hidden if the question is worth zero points.

Previously, the question points have been hidden in the feedback if the points had been set to zero.

Mankro:
I modified the indentation in the HTML code and checked how `field.group_errors` should be used. It is intended to hide errors in one question so that the student would have to answer a group of questions correctly without detailed hints on each question, so the points of one question should be then hidden if `group_errors` is `True`.

PR #48, opened 2019-11-11
Don't show correct answer by default in the questionnaires

alalav1:
Add `show_correct_once` option to exercise configuration to show model answer
after the last submission.


Mankro:
What is the difference between `self.show_correct` and `self.show_correct_once` in the `GradedForm`?

Mankro:
We found the answer: `self.show_correct` is used to render the separate model answer when the "show model answer" button is clicked in A+, whereas `self.show_correct_once` is used to add green OK check marks into the feedback HTML when the student has consumed all submission attempts (the model answer is also shown under text field questions).

Mankro:
I changed the name of the setting and updated the README.

PR #49, opened 2019-11-22
Improved hints

alalav1:
Add regexp based hints in freetext questions with string method. Don't force regexp questions to use regexp compare in hints. Add subdiff directive, to give hints based on difference in solution and submitted answer.

These changes are further explained in the RST-tools. #Aalto-LeTech/a-plus-rst-tools#59


access/types/forms.py
```diff
@@ -753,3 +773,20 @@ def get_randomized_checkbox_attributes(self, i, config, initial, correct,
...
 
+
+def subdiff(value, all_solutions):
+    solutions = all_solutions.split('|')
+    if len(solutions) > 1:
+        matching_parts = ["Multiple correct answers accepted."]
+    else:
+        matching_parts = []
+    for solution in solutions:
+        parts = "Correct parts in your answer: "
```
Mankro: Strings visible to the user must be translated (Finnish/English) -> Django `python manage.py makemessages` etc.

access/types/forms.py
```diff
@@ -753,3 +773,20 @@ def get_randomized_checkbox_attributes(self, i, config, initial, correct,
...
+        matches = difflib.SequenceMatcher(None, value, solution).get_matching_blocks()
+        i = 0
+        for match in matches:
+            parts += '-'*(match.b-i)
```
Mankro: Add whitespace around the operators.

access/types/forms.py
```diff
@@ -587,6 +600,10 @@ def grade_field(self, i, configuration):
                         break
             if add:
                 hints.append(new_hint)
+        # Show matching parts as a feedback, if subdiff is set
+        if methods[0] == 'subdiff' and solution:
```
Mankro: I think it would make sense to generate the subdiff feedback hints in the `grade_text` method since the method returns the hints list anyway. The `grade_text` method would then not need to return any new values.

access/types/forms.py
```diff
@@ -689,11 +706,14 @@ def grade_text(self, configuration, value, hints=None):
             method = "regexp"
         elif "correct" in configuration:
             accept = configuration["correct"]
-        if not accept is None:
+            if method == 'subdiff':
```
Mankro: Can `method` contain modifiers such as `ignorews` here? Then `method == 'subdiff'` would not work.

access/types/forms.py
```diff
@@ -753,3 +773,20 @@ def get_randomized_checkbox_attributes(self, i, config, initial, correct,
         random_attributes['checksum'] = self.samples_hash(random_attributes['sample'])
         return selected_choices, correct_choices, initial_choices, random_attributes
 
+
+def subdiff(value, all_solutions):
```
Mankro: I think we should change the function name `subdiff`. Maybe `get_subdiff_hints` or are there better ideas?

alalav1: The function should be renamed indeed. What about the compare method itself? 'subdiff' might be a little confusing.

Mankro: The compare method can remain `subdiff` since it describes what it does... Some kind of character differences in the string. It is supposed to be a succinct keyword anyway.

access/types/forms.py
```diff
@@ -753,3 +773,20 @@ def get_randomized_checkbox_attributes(self, i, config, initial, correct,
...
+        matching_parts = [_("Multiple correct answers accepted.")]
+    else:
+        matching_parts = []
+    for solution in solutions:
+        parts = _("Correct parts in your answer: ")
+        matches = difflib.SequenceMatcher(None, value, solution).get_matching_blocks()
+        i = 0
+        for match in matches:
+            parts += '-' * (match.b-i)
+            i = match.b+match.size
```
Mankro: Add whitespace around the `+` operator.

locale/fi/LC_MESSAGES/django.po
```diff
@@ -383,6 +383,14 @@ msgstr ""
...
 msgid "Multiple choices are selectable."
 msgstr "Voit valita useita vaihtoehtoja."
+
+#: access/types/forms.py:779
+msgid "Multiple correct answers accepted."
+msgstr "Useita oikeita vastauksia."
+
+#: access/types/forms.py:783
+msgid "Correct parts in your answer: "
+msgstr "Oikeat osat vastauksessasi:"
```
Mankro: Should the Finnish string have space after `:` too?

Mankro:
Don't forget `python manage.py compilemessages` for the translated strings. It creates the binary `.mo` file.

Mankro:
New config values should be added to the `courses/README.md` (section access.types.stdsync.createForm): `subdiff` compare method and the `compare_regexp` under feedback.

Mankro:
Do normal hints in RST work in subdiff questions? The teacher may want to write targeted hints in addition to the automatic difference hint.

alalav1:
> Do normal hints in RST work in subdiff questions? The teacher may want to write targeted hints in addition to the automatic difference hint.

They do work as they should.

Mankro:
`method[7:] if len(method) > 7 else ""`
It is enough and correct to say `method[7:]`. It returns the empty string if the string is too short. List splicing doesn't break even if the list is shorter than implied by the starting index.

PR #50, opened 2019-11-25
Add a no_model option to not show the model answer after the deadline

alalav1:


Mankro:
The name of the setting was changed to `show_model_answer` (from `no_model`).

Mankro:
There is an issue: if the model answer has been previously enabled, then disabling it does not propagate to A+. A+ does not overwrite its setting because it only sees that the `model_answer` field is not set in aplus-json so it does not change the current value at all.

Mankro:
I fixed the issue.

PR #51, opened 2019-12-13
Neutral answers in the model solution

alalav1:
Use question marks in the model answer shown after the last submission to indicate neutral choises, that could have been chosen or not.

### DIFF
--- a/access/types/forms.py
+++ b/access/types/forms.py
@@ -292,6 +292,14 @@ class GradedForm(forms.Form):
                 field.correct = config['model']
             elif config.get('correct', False):
                 field.correct = config['correct']
+            if 'options' in config:
+                neutral = []
+                a = 0
+                for opt in config['options']:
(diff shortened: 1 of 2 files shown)


access/types/forms.py
```diff
@@ -292,6 +292,14 @@ def add_field(self, i, config, field_class, widget_class,
                 field.correct = config['model']
             elif config.get('correct', False):
                 field.correct = config['correct']
+            if 'options' in config:
+                neutral = []
+                a = 0
+                for opt in config['options']:
+                    if opt.get('correct') == 'neutral':
+                        neutral.append(self.option_name(i, opt))
```
Mankro: Either the variable `a` is defined for no reason or `self.option_name(i, opt)` should use `a` instead of `i`.

Mankro: I changed it to use `a`.

PR #58, opened 2020-03-10
Remove code/files related to the Linux chroot sandbox.

jaguarfi:
# Description
- The Linux chroot sandbox is no longer in use. Therefore, files
  and folders related to this implementation are removed.
- The README.md files still have information related to the sandbox.
  However, such content will be removed once the changes are reviewed
  for a reviewer.
- Side note: It is not 100% guaranty that all code related to the
  sandbox has been removed.

# Testing

- [ ] I created new unit tests
- [ ] All unit tests pass
- [x] I have tested manually

- [x] Yes
- [ ] Not yet. I will do it next.
- [ ] Not relevant


access/tasks.py
```diff
@@ -1,5 +1,5 @@
-from celery import shared_task
-from celery.exceptions import SoftTimeLimitExceeded
+from celery import shared_task 
+from celery.exceptions import SoftTimeLimitExceeded
```
raphendyr: Spaces?

access/types/stdasync.py
```diff
@@ -39,7 +39,7 @@
     template_to_str
 from .auth import make_hash, get_uid
 from ..config import ConfigError, DIR
-from .. import tasks
+# from .. import tasks
```
raphendyr: No comments, remove lines

access/types/stdasync.py
```diff
@@ -433,16 +433,6 @@ def _acceptSubmission(request, course, exercise, post_url, sdir):
             r["code"], r["out"], r["err"])
         qlen = 1
 
-    # Queue in celery & rabbitmq for chroot sandbox actions.
-    else:
```
raphendyr: The `container_flag` can still be false, which would silently ignore the submission, so that variable should be removed (also the settings value and so on).

exercises
```diff
@@ -0,0 +1 @@
+/srv/courses
```
raphendyr: Do not add this file to the git, it' for your local testing.

grader/actions.py
```diff
@@ -22,7 +22,7 @@
 from util.files import read_submission_file
 from util.http import get_json
 from util.personalized import user_personal_directory_path, select_generated_exercise_instance
-from util.shell import invoke_script, invoke_sandbox
+from util.shell import invoke_script #, invoke_sandbox
```
raphendyr: remove, no commenting

grader/ajax_key.py
```diff
@@ -0,0 +1,6 @@
+"""
+Automatically generated file.
+This needs to be unique and SECRET. It is also installation specific.
+This file should be included from settings.py
+"""
+AJAX_KEY = 'rr)4qe+#xe7-$lbeaif=+w2ei^c&vc!3rq^*=hw=jbyiekri21'
```
raphendyr: oh. these files can be added to `.gitignore`, in a separate commit. Again, related to local testing.

grader/settings.py
```diff
@@ -133,26 +133,18 @@
...
 CONTAINER_SCRIPT = join(BASE_DIR, "scripts/docker-run.sh")
-CELERY_BROKER = False #'amqp://guest@localhost/'
+CELERY_BROKER = False #'amqp://guest@localhost/' 
 RABBITMQ_MANAGEMENT = { "port": 55672, "password": "guest" }
-CELERY_TASK_LIMIT_SEC = 2 * 60
-CELERY_TASK_KILL_SEC = CELERY_TASK_LIMIT_SEC + 5
-CELERY_AFFINITIES = []
+CELERY_TASK_LIMIT_SEC = 2 * 60 
+CELERY_TASK_KILL_SEC = CELERY_TASK_LIMIT_SEC + 5 
+CELERY_AFFINITIES = []
```
raphendyr: Again some spaces

access/types/stdasync.py
jaguarfi: The logic here is a bit abstract. Nevertheless, I will modify the lines where the `container_flag` is included. Please, review the changes in the next pull request.

.gitignore
```diff
@@ -20,3 +20,18 @@ scripts/chroot_execvp
...
+# symbolic link exercises to /srv/courses
+exercises
+# automatic generated keys
+grader/ajax_key.py
+grader/secret_key.py
+
+# The word "deprecated" was added to the files and directories that use sandbox.
+# The word "deprecated" will help to identify the files that can be removed 
+# definetely.
+*deprecated*
```
Mankro: People normally do not have any files that contain "deprecated" in the file name, so I don't think that the deprecated rule should be in the `.gitignore` file. There is a way to ignore files on your computer without editing the `.gitignore` file that is shared with other developers too: `.git/info/exclude` works like `.gitignore`, but the exclude file is not shared with others (i.e., committed to the repository).

https://git-scm.com/docs/gitignore
https://stackoverflow.com/questions/22906851/when-would-you-use-git-info-exclude-instead-of-gitignore-to-exclude-files

access/types/stdasync.py
```diff
@@ -362,7 +361,7 @@ def _acceptSubmission(request, course, exercise, post_url, sdir):
     '''
     uids = get_uid(request)
     attempt = int(request.GET.get("ordinal_number", 1))
-    container_flag = settings.CONTAINER_MODE and "container" in exercise
+    container_flag = "container" in exercise
 
     # Backup synchronous grading.
     if not container_flag and not settings.CELERY_BROKER:
```
Mankro: This if block for "backup synchronous grading" must be removed. The `runactions` function is only used with the old chroot sandbox. Therefore, this if block does not work with containers.

access/types/stdasync.py
```diff
@@ -336,7 +335,7 @@ def _requireActions(exercise):
     Checks that some actions are set.
     '''
     if "actions" not in exercise or len(exercise["actions"]) == 0:
-        if not settings.CONTAINER_MODE or "container" not in exercise:
+        if not "container" in exercise:
```
Mankro: We don't need the `_requireActions` function at all anymore since we don't use the `actions` field in the exercise config.yaml at all anymore.

access/types/stdasync.py
```diff
@@ -362,7 +361,7 @@ def _acceptSubmission(request, course, exercise, post_url, sdir):
     '''
     uids = get_uid(request)
     attempt = int(request.GET.get("ordinal_number", 1))
-    container_flag = settings.CONTAINER_MODE and "container" in exercise
+    container_flag = "container" in exercise
```
Mankro: The `container_flag` variable looks unnecessary now that only container mode is supported and there is no alternative. The code already calls the `_requireContainer` function before using the `container` field from the exercise config.yaml, which is good. There is no need to make useless if checks (`if container_flag:`) that are always supposed to be True.

If the `container` field is missing from the exercise config.yaml and the `actions` field exists, there could be a clear error to the user (something like "the actions field must be replaced with the container field because the chroot sandbox is no longer supported"). This could be done by raising a `ConfigError` in the `_requireContainer` function.

grader/actions.py
```diff
@@ -22,7 +22,7 @@
 from util.files import read_submission_file
 from util.http import get_json
 from util.personalized import user_personal_directory_path, select_generated_exercise_instance
-from util.shell import invoke_script, invoke_sandbox
+from util.shell import invoke_script
```
Mankro: I think I already told you to remove this whole file `grader/actions.py`. Aren't all these functions used solely with the chroot sandbox? At a quick glance, I still definitively think so.

Mankro: The function names here correspond to the action names used in the actual exercise config.yaml file (in the `actions` field). For example, they always tend to start with the `prepare` action.

grader/settings.py
```diff
@@ -180,10 +172,6 @@
 # Grading action scripts.
 PREPARE_SCRIPT = join(BASE_DIR, "scripts/prepare.sh")
 GITCLONE_SCRIPT = join(BASE_DIR, "scripts/gitclone.sh")
```
Mankro: Shouldn't you also remove `PREPARE_SCRIPT` and `GITCLONE_SCRIPT` since they are only used in the chroot sandbox?

grader/settings.py
```diff
@@ -133,7 +133,7 @@
 ##########################################################################
 CONTAINER_MODE = False
 CONTAINER_SCRIPT = join(BASE_DIR, "scripts/docker-run.sh")
-CELERY_BROKER = False #'amqp://guest@localhost/'
```
Mankro: `CONTAINER_MODE` must be removed since we don't support any other mode any more. Setting it to False would not change anything. `CONTAINER_SCRIPT` is still used and preserved.

I still think that Celery and RabbotMQ settings are not used outside the chroot sandbox, so you need to tell us if there is any real reason to keep them:

* CELERY_BROKER
* RABBITMQ_MANAGEMENT
* CELERY_TASK_LIMIT_SEC
* CELERY_TASK_KILL_SEC
* CELERY_AFFINITIES

You should check if `QUEUE_ALERT_LENGTH` has any real effect with the container grading.

requirements.txt
```diff
@@ -1,7 +1,7 @@
 Django==1.9.7
 raphendyr-django-essentials==1.3.0
 celery==3.1.23
-pyrabbit==1.1.0
+pyrabbit==1.1.0
```
Mankro: No need to add new whitespace to the end of the lines. You must check if the celery package is really still used.

scripts/gitclone.sh
```diff
@@ -10,23 +10,6 @@ RDIR=user-repo
 READ=gitsource
 FILES=()
 
-# Parse arguments.
```
Mankro: Why is this file `scripts/gitclone.sh` not completely deleted? I don't think this is used outside the chroot sandbox. The gitmanager component does not call any script under the scripts directory, right?

scripts/prepare.sh
```diff
@@ -28,30 +28,6 @@ else
   CDIR=courses
 fi
 
-# Parse arguments.
```
Mankro: Why is this file `scripts/prepare.sh` not completely deleted? It is only used in the chroot sandbox, right?

scripts/store_user_files.sh
```diff
@@ -11,20 +11,6 @@ DIR=
 TARGETDIR=
 CP=
 
-# Parse arguments.
```
Mankro: The file `scripts/store_user_files.sh` should also be completely deleted.

Mankro:
You have left some files under the `scripts` directory that I think should be removed because they are only used with the old chroot sandbox. For example, 

* `scripts/_config.sh`
* `scripts/_copy_util.sh`
* `scripts/_tocharset.sh`
* `scripts/pull_request.sh` (this one is rather confusing, but maybe it has been used nowhere in the recent years)

The file `grader/runactions.py` must be removed since it is only used with the old chroot sandbox.

I don't think we need `grader/celery.py` anymore. Where do you think it would be used outside the chroot sandbox?

Mankro:
There are some template files (under `access/templates/access`) that are probably only used with the chroot sandbox because certain template variables are based on the old sandbox `actions` configuration. (Many of the templates in the directory are of course still used.)

Mankro:
The file `util/queue.py` uses Celery, but only to check the queue length, which I think refers to the chroot sandbox grading queue. I don't think Celery is used to send the container grading jobs to Kubernetes. The MOOC-Grader does not anything about the queues inside Kubernetes.

The file `util/affinity.py` might be only used with the old chroot sandbox. Only `access/tasks.py` seems to be using it and the tasks.py module is only used with the chroot sandbox.

The file `util/personalized.py` has some comments that refer to the old chroot sandbox actions. They could be updated with the new container versions. Grading containers can read personalized instance files from the path `/personalized_exercise` if I remember correctly. The user personal directory on the other hand is a strange, extra feature for personalization that we would want to completely remove

.gitignore
jaguarfi: Done!

access/types/stdasync.py
jaguarfi: Done!

access/types/stdasync.py
jaguarfi: Done!

access/types/stdasync.py
jaguarfi: Done!

grader/actions.py
jaguarfi: Yes, I tried to remove the `grader/actions.py` file, but it was causing me some trouble. Now, I removed it completely, and at least the course-template exercises are working fine.

grader/settings.py
jaguarfi: Done!

scripts/prepare.sh
jaguarfi: Done!

scripts/gitclone.sh
jaguarfi: Done!

requirements.txt
jaguarfi: Done!

grader/settings.py
jaguarfi: Done!

scripts/store_user_files.sh
jaguarfi: Done!

access/types/stdasync.py
```diff
@@ -330,20 +321,16 @@ def acceptGeneralForm(request, course, exercise, post_url):
...
-    if "actions" not in exercise or len(exercise["actions"]) == 0:
-        if not settings.CONTAINER_MODE or "container" not in exercise:
-            raise ConfigError("Missing \"actions\" in exercise configuration.")
-
-
 def _requireContainer(exercise):
     c = exercise.get("container", {})
+    a = exercise.get("actions", {})
+    
+    if a:
```
raphendyr: This should be checked only if there is no container configuration. I.e. if there is no container, check if there is actions and edit the warning to state that. If there is neither, then so the default error.

It's very common that actions are left in the configs and this change would break many courses for now. Of course, courses should remove actions. Best is to start raising warnings in build process, which is not easy currently, and kind of is waiting for roman/shepherd.

jaguarfi: Yes, It is true. I will change the `raise ConfigError` for a `LOGGER.warning`

access/management/commands/grade.py
```diff
@@ -50,8 +45,3 @@ def handle(self, *args, **options):
 
             else:
                 raise CommandError("Submit file not found: %s" % (name))
-
-        # Run actions.
-        r = runactions(course, exercise, sdir)
```
Mankro: `runactions` is part of the chroot sandbox, but when you just remove the code here, clearly this `grade` command no longer grades anything. This command should now start the container grader.

access/types/stdasync.py
```diff
@@ -330,20 +321,16 @@ def acceptGeneralForm(request, course, exercise, post_url):
...
-        if not settings.CONTAINER_MODE or "container" not in exercise:
-            raise ConfigError("Missing \"actions\" in exercise configuration.")
-
-
 def _requireContainer(exercise):
     c = exercise.get("container", {})
+    a = exercise.get("actions", {})
+
+    if c and a:
+        LOGGER.warning("The `actions` parameter definded in your cofig.yaml is no longer used by the mooc-grader."
```
Mankro: Typos

* "definded" -> defined
* cofig.yaml -> config.yaml

access/types/stdasync.py
```diff
@@ -390,58 +361,44 @@ def _acceptSubmission(request, course, exercise, post_url, sdir):
...
+    }
+    exercise_extra = {
+        "key": exercise["key"],
+        "title": exercise.get("title", None),
+        "resources": c.get("resources", {}), # Unofficial param, implemented differently later
+        "require_constant_environment": c.get("require_constant_environment", False) # Unofficial param, implemented differently later
+    }
+    if exercise.get("personalized", False):
+        exercise_extra["personalized_exercise"] \
+            = select_generated_exercise_instance(course, exercise, uids, attempt)
```
Mankro: You could add to the git commit message that the personal directory feature has been removed (in addition to the chroot sandbox).

access/types/stdasync.py
```diff
@@ -390,58 +361,44 @@ def _acceptSubmission(request, course, exercise, post_url, sdir):
...
+        c["image"],
+        os.path.join(DIR, course["key"], c["mount"]),
+        sdir,
+        c["cmd"],
+        json.dumps(course_extra),
+        json.dumps(exercise_extra),
+    ])
+    LOGGER.debug("Container order exit=%d out=%s err=%s",
+        r["code"], r["out"], r["err"])
+    qlen = 1
```
Mankro: Is the `qlen` variable used for anything anymore or could it be removed?

Mankro: Wait, I see it in the template context below. Though, the value is always one, but this is not very important to change right now. It is basically not used, but the template is old.

grader/settings.py
```diff
@@ -177,15 +160,6 @@
 # grader stateful, which at least increases the amount of disk space used.
 ENABLE_PERSONAL_DIRECTORIES = False
```
Mankro: You already removed personal directories in the `access/types/stdasync.py` file, so remove this setting `ENABLE_PERSONAL_DIRECTORIES = False` completely too, please.

Mankro:
I wrote new comments and there are still some things to fix.

`util/personalized.py` function `user_personal_directory_path` should be removed since the personal directory feature is removed. Likewise:

* remove `prepare_user_personal_directory`
* `read_user_personal_file` has an if branch for personal directories
* `personalized_template_context` has an if that should be removed

The following things can be left for the future (not this PR):

* Checking old templates under the `access` app and removing the ones that are tied to the chroot sandbox (template variables based on the old actions configuration).
* Celery is still included in requirements.txt, but not used at the moment.
* The docstring at the start of `util/personalized.py` could explain how grading containers use personalized instances.
* `access/types/stdasync.py` has the useless `qlen` variable in the `_acceptSubmission` function.

access/management/commands/grade.py
Mankro: This `grade` command is supposed to grade a submission, so it must start the grader program. Graders can't be run in the old chroot sandbox anymore, so this command must start the grading container by calling `accept.types.stdasync._acceptSubmission`.

Mankro: Note that when `runactions` is called directly here, it runs the exercise grader synchronously. This grade command was intended for testing the grader, not for real production use. We can't really run container graders synchronously, so there might not be a way to receive the output of the container (feedback and points) here. Without the output, running the grader is rather pointless, so the whole command might become obsolete.

access/types/stdasync.py
jaguarfi: I think I will handle the removal of the personal directory in a different PR. It seems like there are a couple of files that need to be modified/removed.

Mankro:
It is hard to see what has been changed when you squash and force-push after every small change. I guess you fixed the typos and removed the `access/management/commands/grade.py` file today. Anyway, this should be finished now since we can leave the removal of the personal directory to another PR.

@raphendyr Shall we merge this now? It should be good now. My comments in the previous message can be handled after this PR.

jaguarfi:
@Mankro Yes, I applied those changes. Btw, I think, I have to read how to contribute to opensource projects. I mean, I do not know the workflow really well. For example, Should I create a different PR  every time that I change something? or should I create a new commit with those changes?

Mankro:
We haven't yet written the contributor's guide for the A+ project, so basically I don't know the "right" workflow either. Anyway, at first I would say this:

* You create new PRs for new features and fixes that logically belong together. So, one PR for one feature. The contents of one PR should constitute a logical set of changes for the whole feature. A pull request does not always add a new feature, since there are also bug fixes and other necessary changes without adding new features.
* You modify the PR by pushing new commits when you fix bugs and other issues that have been pointed out in the code review and testing.
* You should clean up the git history before creating the PR. While the PR is being reviewed, you make changes in new commits and do not squash so that reviewers can easily follow the latest changes (the newest commits). At the end when everything has been polished, you clean up the git history again (git squash etc.) and force-push.
* There can be more than one commit. Each commit should constitute a sensible set of changes that somewhat belong together. The code should usually always "work" after each commit, so don't leave the code in a completely broken state after a commit (syntax errors, immediate crashes when executing the code). Of course, not every detail can work after a single commit when large features are split into many commits, but there are usually logical subcomponents or such.

PR #77, opened 2020-07-22
Add description of the `model_files` to the courses/README.md

jaguarfi:
# Description

The `model_files` configuration exists in the source code, but
there is no mention of that specific configuration value in the
courses/README.md file. Therefore, a brief description has been
added

### DIFF
--- a/courses/README.md
+++ b/courses/README.md
@@ -154,6 +154,9 @@ Durations are given in (int)(unit), where units are y, m, d, h or w.
 		defines how many times the student may submit before the personalized exercise is
 		regenerated (the exercise instance is changed to another one). If unset,
 		the exercise is never regenerated.
+	* `model_files`: It is a list of model answers that are available only after the deadline has passed.
+		The `model_files` take file paths as input. These paths are relative to the root of the course repository,
+		e.g., `exercises/hello_world/model.py`.
 
 	Rest of the attributes are exercise type specific.
 


raphendyr:
The code first checks field `model_answer` then `model_files` and finally, if both of those were missing, it handles `access.types.stdsync.createForm` differently. Hence, both `model_answer` and `model_files` works with all `view_type`s, not just `postFiles`.

jaguarfi:
> Hence, both `model_answer` and `model_files `works with all view_types, not just `postFiles`.

I only added this to the  `access.types.stdasync.acceptFiles` not the `postFiles`. Anycase, I did not want to add this to the other `access.types` because I only tested this in `access.types.stdasync.acceptFiles`. **But**, If you are completely sure that the code behaves as expected I can add the `model_files` to all the other `access.types` excluding `access.types.stdasync.acceptPost`.

I did not add `model_answer` to the description because I could not make it work. :thinking:

raphendyr:
The code applies to all view_types https://github.com/apluslms/mooc-grader/blob/master/util/export.py#L75

jaguarfi:
My bad!!! :facepalm: . I meant `access.types.stdsync.createForm` instead of   `access.types.stdasync.acceptFiles. I will add the value `model_files` to all the `access.types` except `access.types.stdsync.createForm`.

raphendyr:
Or you could add it under `course_key/exercise_key.[json|yaml]` as that is the common list of options, which are not specific to a view_type.

jaguarfi:
but what about `access.types.stdsync.createForm`?

raphendyr:
The `model_files` has priority over `view_type == createForm`.

jaguarfi:
Yes, I meant all execpt `createForm`
> I will add the value model_files to all the `access.types` except `access.types.stdsync.createForm`

I would not like to add it to `course_key/exercise_key.[json|yaml]` because `createForm` does not use the `model_files` value. Therefore, it makes more sense to add the `model_files` value to each  `access.types` that accept the `model_files` value rather than adding it to the general configuration `course_key/exercise_key.[json|yaml]`.

raphendyr:
> because createForm does not use the model_files

The code does use the `model_files  for that too, `createForm` just has a default if that is missing. However, using that for `createForm` makes no sense. Maybe if the `createForm` is used with a container grader (that is possible too).

jaguarfi:
Ok, I will add the values `model_files` to the `course_key/exercise_key.[json|yaml]`.

Mankro:
It would be better to more clearly describe what kind of values should be given to `model_files`. The list of model answers contains *file paths* starting from the root of the course repo, e.g., `exercises/hello_world/model.py`.

courses/README.md
```diff
@@ -154,8 +154,10 @@ Durations are given in (int)(unit), where units are y, m, d, h or w.
 		defines how many times the student may submit before the personalized exercise is
 		regenerated (the exercise instance is changed to another one). If unset,
 		the exercise is never regenerated.
+	* `model_files`: list of model answers. The model answers are available after the deadline has passed.
```
Mankro, line 157: Like I wrote in the discussion, this description could emphasize that the model answers are file paths.

raphendyr: Specially, that they are paths, which **do not** start with `/`. Relative paths typically are relative to the file they are in, and not some random location. This is an example of a bad mistake, which can't be fixed.

Mankro (approved):

PR #78, opened 2020-08-01
Questionnaire feedback summary - changes in mooc-grader

saarasat:
# Description

~~Preferably not to be merged before: https://github.com/apluslms/mooc-grader/pull/76~~

Related to the changes in: https://github.com/apluslms/a-plus/pull/637

Fixes: https://github.com/apluslms/a-plus/issues/592 

Related also to solving: https://github.com/apluslms/a-plus/issues/593 and https://github.com/apluslms/a-plus/issues/591 

**1. Implements auto_id for questionnaire fields.**
- The auto_id contains a random string as a separating factor for different fields. 
- This has been done for the cases where there are multiple questionnaires on the same page, this separates the fields of those questionnaires for better accessibility, and also makes handling errors on the front-end easier. 

**2. Removes the required-attr addition from questionnaire widgets** 
- Now that Django has been updated, the widgets no longer needed the html-attribute of required set. So validation errors are handled all by Django for consistency. 

**3. Presents a validation error summary for the questionnaire:** 
- Previously the validation errors for questionnaires were mixed in with the errors presented of wrong answers. Those two were separated in: https://github.com/apluslms/mooc-grader/pull/76. This continues on that, by adding a validation error summary at the top of the questionnaire, which is recommended eg. in [here](https://design-system.service.gov.uk/patterns/validation/). If there are errors with the form and with the users input in those, those are showed.
- The html-template of [_error_summary.html](https://github.com/apluslms/mooc-grader/compare/master...saarasat:quiz-feedback-summary?expand=1#diff-88f59fe9e07c41e9751ba71bc6060a39) could be used later on for other exercise-types and their errors for continuity and easier handling of all mooc-grader errors, for example at:
https://github.com/apluslms/mooc-grader/blob/master/access/templates/access/exercise_frame.html
https://github.com/apluslms/mooc-grader/blob/master/access/templates/access/active_element_default.html
https://github.com/apluslms/mooc-grader/blob/master/access/templates/access/async_accepted.html
https://github.com/apluslms/mooc-grader/blob/master/access/templates/access/exercise_config_error.html

**4. Presents an initial style for possible wider user for (validation) error summaries** 
- The summary has been styled and worked on based on @mtthwhggns initial design. 
- **The design could be still easily worked on, if deemed somehow unfit. Or if the style is accepted, it could be adopted as a general replacement for the current bootstrap alert-danger, at least for the exercise-error** 

**5. Removes the duplicate information of points from questionnaire** 
- Additionally as the feedback for the student about wrong and right answers were improved in https://github.com/apluslms/mooc-grader/pull/76, this removes the now unnecessary duplicate information regarding the points of a questionnaire. If the teacher has set an accepted-message to be shown after a submission is accepted (regardless if the question is right or wrong), that will be shown with the default-alert.

**This pull request DOES NOT:** 
- Work on the exercises in general. This is just a proof of concept for the questionnaire on the error summary, and it could be then spread for wider use within the exercises. 

**Error summary. The error-message showing below the field is improved in https://github.com/apluslms/mooc-grader/pull/76**
image

**Accepted message**
image 

# Testing

The changes have been tested manually and the existing automatic tests pass.

*Don't forget to update the documentation if it is relevant for this PR.*

### DIFF
--- a/access/templates/access/_error_summary.html
+++ b/access/templates/access/_error_summary.html
@@ -1,0 +1,18 @@
+{% load i18n %}
+
+
+<div class="error-summary">
+	<div class="error-summary__text">
+		<h4>{% trans "Submission failed" %}</h4>
+		<p>{% trans "Check your answers before trying again:" %}</p>
+		{% for field in result.form %}
+			{% if field.errors %}
+				<p>{{ field.label }}:
+				{% for error in field.errors %}
+					{{ error }}
+				{% endfor %}
+				</p>
+			{% endif %}
+		{% endfor %}
+	</div>
+</div>
--- a/access/templates/access/create_form_default.html
+++ b/access/templates/access/create_form_default.html
@@ -13,7 +13,7 @@
 			<p>{{ exercise.accepted_message|safe }}</p>
 		{% endif %}
 		<p>
-			{% blocktrans %}
+			{% blocktrans trimmed %}
 				The whole section within the black line is marked as being
 				incorrect, if one of the questions incorrect.
 			{% endblocktrans %}
--- a/access/templates/access/create_form_default.html
+++ b/access/templates/access/create_form_default.html
@@ -3,32 +3,10 @@
 
 {% block exercise %}
 
-{% if result.accepted %}
-<div class="alert {% if result.error_fields %}alert-danger{% else %}alert-success{% endif %}">
-<p>
-    {% if exercise.accepted_message %}
-    {{ exercise.accepted_message|safe }}
-    {% else %}
-    {% blocktrans with points=result.points max_points=exercise.max_points %}
-    You got {{ points }}/{{ max_points }} points from this questionnaire.
-    {% endblocktrans %}
-    {% endif %}
-</p>
-<p>
-    {% if result.form.group_errors %}
-    {% blocktrans %}
-    Note that the whole section of questions is marked as being incorrect if
-    one of the questions in that section is incorrect.
-    {% endblocktrans %}
-    {% endif %}
-    {% if not result.form.disabled and result.points < exercise.max_points %}
-    {% blocktrans %}
-    You can change your answers and submit them for regrading.
-    However, notice the number of allowed submissions left!
-    {% endblocktrans %}
-    {% endif %}
-</p>
-</div>
+{% if result.accepted and exercise.accepted_message %}
+  <div class="alert alert-default">
+    <p>{{ exercise.accepted_message|safe }}</p>
+  </div>
 {% endif %}
 
 {% if exercise.instructions %}
--- a/access/templates/access/create_form_default.html
+++ b/access/templates/access/create_form_default.html
@@ -7,6 +7,18 @@
 	<div class="alert alert-default">
 		<p>{{ exercise.accepted_message|safe }}</p>
 	</div>
+{% elif result.accepted and result.form.group_errors %}
+	<div class="alert alert-default">
+		{% if exercise.accepted_message|safe %}
+			<p>{{ exercise.accepted_message|safe }}</p>
+		{% endif %}
+		<p>
+			{% blocktrans %}
+				The whole section within the black line is marked as being
+				incorrect, if one of the questions incorrect.
+			{% endblocktrans %}
+		</p>
+	</div>
 {% elif result.form.errors %}
 	{% include "access/_error_summary.html" %}
 {% endif %}
--- a/access/templates/access/create_form_default.html
+++ b/access/templates/access/create_form_default.html
@@ -4,18 +4,20 @@
 {% block exercise %}
 
 {% if result.accepted and exercise.accepted_message %}
-  <div class="alert alert-default">
-    <p>{{ exercise.accepted_message|safe }}</p>
-  </div>
+	<div class="alert alert-default">
+		<p>{{ exercise.accepted_message|safe }}</p>
+	</div>
+{% elif result.form.errors %}
+	{% include "access/_error_summary.html" %}
 {% endif %}
 
 {% if exercise.instructions %}
 <div class="instructions">
-  {{ exercise.instructions|safe }}
+	{{ exercise.instructions|safe }}
 </div>
 {% endif %}
 {% if exercise.instructions_file %}
-  {% include exercise.instructions_file %}
+	{% include exercise.instructions_file %}
 {% endif %}
 
 {% include 'access/graded_form.html' %}
--- a/access/types/forms.py
+++ b/access/types/forms.py
@@ -238,19 +238,6 @@ class GradedForm(forms.Form):
         if self.disabled:
             args['widget'].attrs['disabled'] = 'disabled'
 
-        if args['required'] and (
-            widget_class is not forms.CheckboxSelectMultiple or len(choices) == 1
-        ):
-            # The required HTML attribute should not be set to a question with
-            # multiple checkboxes since the browser would require that the user
-            # selects every checkbox.
-            # Note: newer Django versions include the required HTML attribute
-            # automatically when the "required" parameter is given to the Field
-            # constructor (like is done here with the args variable).
-            # When Django is upgraded, we don't need to set the "required" attr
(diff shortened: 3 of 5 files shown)


access/types/forms.py
```diff
@@ -38,6 +38,13 @@ def __init__(self, *args, **kwargs):
             self.show_correct_once = False
         self.request = kwargs.pop('request') if 'request' in kwargs else None
         kwargs['label_suffix'] = ''
+
+        # Set form-field ids based on exercise-key
+        exercise_key = self.exercise.get("key", False)
+        if not exercise_key:
+            exercise_key = self.exercise.get("title", "")
+        kwargs.setdefault("auto_id", "{}_%s".format(exercise_key))
```
raphendyr, line 46: When the same questionnaire is open in the exercise frame and in a modal dialog (past submission), then these ids will be duplicated. Hence, it would be beneficial to add a bit of randomness to it. Technically ordinar number should work too, but there can be cases where that assumption breaks.

I started using format `exercise-{{ form_random_id }}-field-{{ field.name }}` for field ids and I created form html with

```django
<form id="exercise-{{ form_random_id }}-form" data-form-nonce="{{ form_random_id }}">
```

where

```python
from django.utils.crypto import get_random_string
form_random_id = get_random_string(length=8)
```

Thus, maybe?

```python
random_id = get_random_string(length=8)
kwargs.setdefault("auto_id", "exercise-{}-field-%s".format(random_id))
super().__init__(*args, **kwargs)
self.form_id = "exercise-{}-form".format(random_id)
self.form_nonce = random_id
```

and then you can use `form.form_id` and `form.form_nonce` in the html template.

access/types/forms.py
```diff
@@ -146,7 +153,7 @@ def __init__(self, *args, **kwargs):
                     fi.group_errors = group_errors
 
                 if j == 0:
-                    f[0].open_set = self.group_name(g)
+                    f[0].open_set = "{}_{}".format(exercise_key, self.group_name(g))
```
raphendyr, line 156: I guess this could be `"exercise-{}-set-{}".format(self.form_nonce, self.group_name(g))`

access/types/forms.py
saarasat: Oh, good to know. I followed the example in mooc-jutut, but did not realize the specificity of questionnaire in this case. I now implemented this version, works nicely :)

access/templates/access/_error_summary.html
```diff
@@ -0,0 +1,20 @@
+{% load i18n %}
+
+
+<div class="error-summary">
+	<div class="background">
+		<div class="text-container">
```
raphendyr, line 6: Does this really require three nested divs to get correct positioning etc? Seems strange as I would expect that two should be sufficient, but maybe I'm wrong.

saarasat: This was done based on earlier suggestions in the other repo, but two will do as well, so the change has been made. Just let me know then if there is something else here that would require changing or thinking about. Or alternatively whether this pr has become squashable :)

access/types/forms.py
```diff
@@ -230,19 +238,6 @@ def add_field(self, i, config, field_class, widget_class,
         if self.disabled:
             args['widget'].attrs['disabled'] = 'disabled'
 
-        if args['required'] and (
-            widget_class is not forms.CheckboxSelectMultiple or len(choices) == 1
-        ):
-            # The required HTML attribute should not be set to a question with
-            # multiple checkboxes since the browser would require that the user
-            # selects every checkbox.
-            # Note: newer Django versions include the required HTML attribute
```
Mankro, line 239: I can't see anywhere that the Django version would have been upgraded in the MOOC-Grader. If it has not been upgraded, then this if branch shouldn't be deleted, right?

saarasat: Hmm.. then I don't know what has happened in between writing those comments and now. Since it seems to work as expected now without that extra attribute set. 

The browser does not require the user to check every checkbox, just one. And the validation happens when the form is being assessed, so the appearance is correct and consistent for checkboxes without answers:

image

raphendyr: Verify that the html attribute `required` is not defined for input checkbox.

access/templates/access/_error_summary.html
```diff
@@ -0,0 +1,18 @@
...
+
+
+<div class="error-summary">
+	<div class="error-summary__text">
+		<h4>{% trans "Submission failed" %}</h4>
+		<p>{% trans "Check your answers before trying again:" %}</p>
+		{% for field in result.form %}
+			{% if field.errors %}
+				<p>{{ field.label }}:
+				{% for error in field.errors %}
```
Mankro, line 11: If one field really has multiple errors, are the errors just concatenated inside one `<p>` here? Is there at least some space between them? Did you try it to see how it looks? Hmm, how could we set up multiple errors for testing?

access/templates/access/create_form_default.html
```diff
@@ -3,41 +3,21 @@
...
-    You got {{ points }}/{{ max_points }} points from this questionnaire.
-    {% endblocktrans %}
-    {% endif %}
-</p>
-<p>
-    {% if result.form.group_errors %}
-    {% blocktrans %}
-    Note that the whole section of questions is marked as being incorrect if
-    one of the questions in that section is incorrect.
-    {% endblocktrans %}
```
Mankro, line 22: Was there a reason for removing this warning about question groups? Does the UI otherwise imply clearly that grouped questions require correct answers on all questions? (This setting is not often used these days since the RST questionnaire format doesn't even have any syntax for defining these question groups?)

access/types/forms.py
saarasat: Doesn't seem to be: 
image

saarasat: The current versions of Chrome and Firefox I have work perfectly

access/templates/access/_error_summary.html
saarasat: That is a good point. I only tested the outlook with multiple erroneous fields, but not if the same field has multiple errors. I checked it now by adding manually another error within the forms.py, as I couldn't think of away to introduce multiple errors at the same time with the current validations we have on those fields. I guess those cases luckily are quite rare. As if the format is wrong, only one error is raised saying to insert the correct format (eg. a number). And then again, if something is inserted, there is no error for the missing field.

Looks like there is always the dot and one space before the error. So the errors are listed automatically pretty nicely. :) I think this works as it is for those cases? 

image

Mankro: Yes, that is good enough :+1:

access/templates/access/create_form_default.html
saarasat: Good point. Maybe it indeed wasn't clear enough. But on the other hand I don't think it was clear enough before either. It was like this before:

image
 
But the red danger alert was removed in the previous pull request. I now implemented it somewhat back, but instead of using the red alert, I used the black line of the `alert alert-default`, and indicated that whichever questions are within those lines and form the question group that has errors, are marked incorrect. It is not ideal, but better:

image

And like you said, since there is no directive for the group, I think (and hope) that it was not used much before.

saarasat: This was the earlier implementation alongside with the message you pointed out:
https://github.com/apluslms/mooc-grader/blob/fcdfd7fee9e3993d2f26d625afe95538cd4b1359/access/templates/access/graded_form.html#L5

Mankro: Great, now it's better!

access/types/forms.py
Mankro: Accessibility guidelines don't really recommended using the browser built-in validation with the `required` HTML attribute, so even if the attribute is not included now when the code is removed, we can accept that.

Mankro:
Did you rebase already? I could accept this now that the latest comments and fixes have been finished.

saarasat:
> Did you rebase already? I could accept this now that the latest comments and fixes have been finished.

Yes! I rebased and checked that everything works while doing the previous changes. So should be good now :)

Issue #55, opened 2020-03-03
Questionnaire model answer shows neutral choices as incorrect


jaguarfi:
> **Migrated from Trello**

When you press the A+ "show model answer" button in a questionnaire exercise, it shows the neutral choices of checkbox questions as incorrect. There is no visual separation between incorrect and neutral choices.

annirytkonen:
this should be moved to MOOC grader

Mankro:
I had a feeling that this was fixed in December 2019, but it must be checked.

PR #80, opened 2020-08-11
Neutral options

saarasat:
# Description

Fixes: https://github.com/apluslms/mooc-grader/issues/55 

Neutral options of checkbox-type questions have been added to the model answer in questionnaire. Previously they only showed briefly at the moment when the student reached maximum amount of submissions if the “reveal-model-at-max-submissions" was set to true. 

At the same time, the role of these two different model answers have been cleared up a bit and pushed towards the other [general exercise configuration](https://github.com/apluslms/mooc-grader/tree/master/courses#accesstypesstdsynccreateform), by renaming the form attributes for the model answers.

The only visible change to appearance, is the indication of neutral answers and hiding the option-related hints, when checking the model answer:

image

# Testing 

Has been tested with: 

1. Normal checkboxes without neutral option(s) 
2. Normal checkboxes with neutral option(s) 
3. Randomized checkboxes with neutral option(s) 
4. Randomized checkboxes without neutral option(s) 
5. Questionnaires with other types of questions at the same time as checkbox-questions (radio-, text- and dropdown-fields)
6. All the above, while there are general and option-related hints available 

Existing automatic tests also pass.

Not really relevant

### DIFF
--- a/access/types/forms.py
+++ b/access/types/forms.py
@@ -34,10 +34,10 @@ class GradedForm(forms.Form):
         if "exercise" not in kwargs:
             raise ConfigError("Missing exercise configuration from form arguments.")
         self.exercise = kwargs.pop("exercise")
-        self.show_correct = kwargs.pop('show_correct') if 'show_correct' in kwargs else False
-        self.show_correct_once = kwargs.pop('show_correct_once') if 'show_correct_once' in kwargs else False
+        self.model_answer = kwargs.pop('model_answer') if 'model_answer' in kwargs else False
+        self.reveal_correct = kwargs.pop('reveal_correct') if 'reveal_correct' in kwargs else False
         if not self.exercise.get('reveal_model_at_max_submissions', False):
-            self.show_correct_once = False
+            self.reveal_correct = False
         self.request = kwargs.pop('request') if 'request' in kwargs else None
         kwargs['label_suffix'] = ''
 
@@ -52,7 +52,7 @@ class GradedForm(forms.Form):
 
         self.form_id = "exercise-{}-form".format(random_id)
         self.form_nonce = random_id
-        self.disabled = self.show_correct
+        self.disabled = self.model_answer
         self.randomized = False
         self.rng = random.Random()
         self.multipart = False
@@ -199,7 +199,7 @@ class GradedForm(forms.Form):
         choices, initial, correct = self.create_choices(config)
         for row in config.get('rows', []):
 
-            if self.show_correct:
+            if self.model_answer:
                 correct = []
                 corr = self.row_options(config, row)['options']
                 for i,choice in enumerate(choices):
@@ -252,7 +252,7 @@ class GradedForm(forms.Form):
                 )
             args['choices'] = selected_choices
 
-        if self.show_correct:
+        if self.model_answer:
             if correct:
                 args['initial'] = correct_choices if multiple else correct_choices[0]
             elif config.get('model', False):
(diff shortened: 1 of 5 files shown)


access/types/forms.py
```diff
@@ -33,18 +33,18 @@ def __init__(self, *args, **kwargs):
         if "exercise" not in kwargs:
             raise ConfigError("Missing exercise configuration from form arguments.")
         self.exercise = kwargs.pop("exercise")
-        self.show_correct = kwargs.pop('show_correct') if 'show_correct' in kwargs else False
-        self.show_correct_once = kwargs.pop('show_correct_once') if 'show_correct_once' in kwargs else False
+        self.model_answer = kwargs.pop('model_answer') if 'model_answer' in kwargs else False
+        self.reveal_correct = kwargs.pop('reveal_correct') if 'reveal_correct' in kwargs else False
```
raphendyr, line 37: `self.reveal_correct = kwargs.pop('reveal_correct', False)` should be sufficient

access/types/stdsync.py
```diff
@@ -151,11 +151,12 @@ def createForm(request, course, exercise, post_url):
...
 
 def createFormModel(request, course, exercise, parameter):
-    form = GradedForm(None, exercise=exercise, show_correct=True)
+    form = GradedForm(None, exercise=exercise, model_answer=True)
     form.bind_initial()
     points,error_groups,error_fields = form.grade()
     result = { "form": form, "accepted": True, "points": points,
-        "error_groups": error_groups, "error_fields": error_fields }
+        "error_groups": error_groups, "error_fields": error_fields,
+        "model_answer": True }
```
raphendyr, line 159: This would be a good time to move every key on it's own line.

access/types/forms.py
saarasat: Very true, I changed those.

access/types/forms.py
```diff
@@ -323,26 +320,29 @@ def create_more(self, configuration):
...
+        neutral = []
         if "options" in configuration:
             i = 0
             for opt in configuration["options"]:
                 label = opt.get('label', "")
                 value = self.option_name(i, opt)
                 choices.append((value, mark_safe(label)))
+                if opt.get('correct') == "neutral":
+                    neutral.append(value)
                 if opt.get('correct', False) is True:
```
Mankro, line 338: `elif` could be used since `opt.get('correct')` can not be both `"neutral"` and `True` at the same time. Now, two `if` conditions are always checked.

locale/fi/LC_MESSAGES/django.po
```diff
@@ -332,21 +332,25 @@ msgstr[1] ""
 "\n"
 "%(points)s pistettä"
 
-#: access/templates/access/graded_form_feedback.html:15
+#: access/templates/access/graded_form.html:134
+msgid "neutral option which does not affect grading"
+msgstr "vaihtoehto ei vaikuta kokonaispisteisiin"
```
Mankro, line 337: I wonder if the Finnish string should include "neutral": "neutraali vaihtoehto ei vaikuta kokonaispisteisiin".

access/types/forms.py
saarasat: You are correct! I changed that now

locale/fi/LC_MESSAGES/django.po
saarasat: Added that as well, made sense to me.

saarasat:
This one is also now squashed, tested, and rebased on top of the current master (https://github.com/apluslms/mooc-grader/commit/0df9b3b197171396ec97dbab2b3481b6a494422a). Also translations have been checked.

PR #81, opened 2020-08-19
Disable points and feedback from enrollment questionnaire

saarasat:
The status-text "Correct!" and the points badges (0/0 for all questions, regardless of the answer) were shown for enrollment questionnaires too. This removes them as unnecessary feedback. The green alert of "you got 0/0 points for this questionnaire" was already removed in an earlier pull request. 

**Before changes** 

Screenshot from 2020-08-19 09-07-34

**After changes** 

image 

If there is an required answer missing from the questionnaire, that is displayed now also better due to the earlier pull requests:

image

image

Doesn't require updates. 

### DIFF
--- a/access/templates/access/graded_form.html
+++ b/access/templates/access/graded_form.html
@@ -37,7 +37,7 @@
 			{% endif %}
 
 			{% if not field.field.group_errors %}
-			{% if result.accepted %}
+			{% if result.accepted and not result.form.enrollment_exercise %}
 			<!-- Points badge -->
 				{% if not exercise.feedback or field.field.points > 0 %}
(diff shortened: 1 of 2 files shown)


raphendyr:
Would it make sense to remove the points tooltip always if the max points are 0 for an item or if there is no correct/invalid answers?

Mankro (approved):

Mankro:
> Would it make sense to remove the points tooltip always if the max points are 0 for an item or if there is no correct/invalid answers?

Maybe, but we should first consider different use cases. We can't do it now since we are running out of time, so I would leave that idea for the distant future.

I could merge this if there are no objections.

raphendyr (approved):
Yes. Can you @Mankro create an issue to remind to revisit the topic.

Mankro:
> Yes. Can you @Mankro create an issue to remind to revisit the topic.

Created #86

PR #82, opened 2020-08-19
Change style of list-items in error summary

saarasat:
Changes the style of the items in the error summary,
adds emphasis on the field label, and presents the errors
as ul-element, instead of individual p-elements.

Related to the changes: https://github.com/apluslms/a-plus/pull/658
And in: https://github.com/apluslms/a-plus-design-system/pull/9

# Testing

Testing of the outlook has been done in: https://github.com/apluslms/a-plus/pull/658

### DIFF
--- a/access/templates/access/_error_summary.html
+++ b/access/templates/access/_error_summary.html
@@ -5,14 +5,17 @@
 	<div class="error-summary__text">
 		<h4>{% trans "Submission failed" %}</h4>
 		<p>{% trans "Check your answers before trying again:" %}</p>
+		<ul>
 		{% for field in result.form %}
 			{% if field.errors %}
-				<p>{{ field.label }}:
-				{% for error in field.errors %}
-					{{ error }}
-				{% endfor %}
-				</p>
+				<li>
+					<strong>{{ field.label }}: </strong>
+					{% for error in field.errors %}
+						{{ error }}
+					{% endfor %}
+				</li>		
 			{% endif %}
 		{% endfor %}
+		</ul>
 	</div>
 </div>


access/templates/access/_error_summary.html
```diff
@@ -5,14 +5,17 @@
...
 		{% for field in result.form %}
 			{% if field.errors %}
-				<p>{{ field.label }}:
-				{% for error in field.errors %}
-					{{ error }}
-				{% endfor %}
-				</p>
+					<li>
+						<strong>{{ field.label }}: </strong>
+						{% for error in field.errors %}
```
Mankro, line 13: Now that you are using a html list here, would it make sense to show each field error as a separate list item? Now you have one list item per field, but the field may have multiple errors in the same list item, though it is very rare. What do you think?

Mankro (approved):

access/templates/access/_error_summary.html
saarasat: That crossed my mind, but I kind of like it better, when they are as a list by field, so that the errors of the same field are there one after the other. I am not sure what is the convention here though? That could of course be changed still :)

Mankro: I think either way is fine, so it can be left as is.

access/templates/access/_error_summary.html
```diff
@@ -5,14 +5,17 @@
 	<div class="error-summary__text">
 		<h4>{% trans "Submission failed" %}</h4>
 		<p>{% trans "Check your answers before trying again:" %}</p>
+		<ul>
 		{% for field in result.form %}
```
raphendyr, line 9: The `for` and the following `if` is not indented. Alternatively,  the `<li>` is indented too much.

saarasat: You are correct, this was fixed

raphendyr (approved):

Issue #66, opened 2020-04-16
Implement drone.io plugin type container for uploading course data to the API


jaguarfi:
>**Migrated from Trello**

Requires API: https://trello.com/c/fEUXtsE4

Using the container would look something like this:

```dockerfile
docker run --rm \
  -w /data/ \
  -v "$(pwd):/data/:ro" \
  -e PLUGIN_API=http://172.16.0.1:8080/api/ \
  -e PLUGIN_API_TOKEN=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJkZWZfY291cnNlIn0.SnjF78daKOfCwqMuljTQvVMg3C5HyvdAknKF2eVjdwo \
  apluslms/deploy-mooc-grader
```

Where `$(pwd)` should yield course directory, which has been build using `roman` (or `docker-compile.sh`).

Tools to consider:

* apluslms-yamlidator for parsing and validating yaml files (or Document class if those files should be updated too)
  * package is part of https://github.com/apluslms/roman/
* some existing work on porting yaml parsing from mooc-grader to external library
  * https://github.com/ematan/roman/tree/aplus-json/apluslms_roman
  * specially the files `test.py`, `access.py`, `dataset.py`, `parser.py` and schemas

Extra info:

* drone.io is a CI tool using pipelines composed of docker containers and commands or "plugins"
* Goal is for roman to support drone.io plugin interface. You might get some idea from the drone docs: https://docs.drone.io/plugins/

----
**Comments**

>**Jaakko Kantojärvi:**
apluslms-yamlidator is going to PyPi in near future, so it can be installed from there at some point.

>Use target branch c`loud-dev` when ever creating PR for the card.

Issue #63, opened 2020-04-09
MOOC-grader leaks secret chapters to the web


jaguarfi:
>**Migrated from Trello**

All chapter content is compiled in the MOOC-grader and served to the web from the `_build/html` directory. However, since some chapters have restricted visibility in A+, it makes no sense that their content is served to the public web from the MOOC-grader. The content should not be available to the web in the MOOC-grader since it is supposed to be served via A+ and follow access restrictions in A+ (where the user has been authenticated and restrictions can be applied).

Mankro:
Blocked by: the Shepherd project and/or a new solution for serving chapter HTML files and static files to A+.

MOOC-Grader won't be fixed, but the chapter HTML files and the gitmanager will be removed from it.

Mankro:
The new gitmanager project should solve this issue.

Mankro:
Closing because the new Git Manager fixed this.
https://github.com/apluslms/gitmanager/blob/365b1c7d9fbd40f8503d7addfa3ae6e2062bbfa1/access/views.py#L101

Issue #57, opened 2020-03-09
gitmanager: Define env value STATIC_CONTENT_HOST


raphendyr:
Define `STATIC_CONTENT_HOST` in `gitmanager/cron_pull_build.sh`, so that A+ and RST build can start to rely on it. Currently, both of those have hacks to go around the issue. Main goal is to ease the migration by allowing us to start developing tools, while still using the old deployment tool (gitmanager).

In a long run, static material will be moved to private or public CDN, thus it will not be in a same place as the course chapter html or exercises.

raphendyr:
Related to #31

Mankro:
The new gitmanager service in A+ v1.12 defines `STATIC_CONTENT_HOST`.

https://github.com/apluslms/gitmanager/blob/912a7fceff8b232445bbe6cb6d5791f2f4daccf3/builder/builder.py#L112

Issue #31, opened 2018-08-23
Support rewriting of static_dir urls


piehei:
Currently if you try to show an image (or link to a downloadable file or whatever) in an erercise_config.yaml based exercise instructions, you have to know the hard-coded *actual* url where the image resides at the production server, and the image file of course has to be copied (or symlinked to) the static directory of the course. This affects all the cases where you want to link to some file in the course static dir.

If you cannot completely adopt the RST workflow, you have this problem.

This comes with problems:
1. when you run local dev environment (course-templates), mooc-grader serves your course from /default which makes your static dir /static/default
2. when you deploy your course to production, your course key changes from default to something, thus breaking links that worked in local dev environment
3. none of the usual ways of linking work, you cannot do "./static/blah.jpg", "/static/blah.jpg", "../static/blah.jpg" etc. because the static path depends *solely* on the dir where the course files reside on the production server

Or am I missing something? (Totally could be!!!)

One way to solve:
rewrite all urls in exercise_config.yaml so that they point to the correct static path on the server, eg. if you put "/static/my_file.png" in some exercise_config.yaml, it would be rewritten to "/static/COURSE_KEY/my_file.png"; this solution takes a couple of lines of code only and I have it ready since I played around with it a bit (I implemented this in util/templates.py).

What do you think?

:)

raphendyr:
Wait for roman

Mankro:
This issue is old and the handling of links has changed a little over the years, though the problem presented in this issue is still relevant.

A-plus-rst-tools makes some changes to the URLs so that they work in the A+ frontend. This affects exercise instructions HTML too. Manually written HTML in exercise config.yaml files may use the same format that a-plus-rst-tools output. For example, an image tag:

`<img alt="Aplus logo" data-aplus-path="/static/{course}" src="../_images/apluslogo.png"/>`




The new gitmanager service in A+ v1.12 defines `STATIC_CONTENT_HOST` for the build container. A-plus-rst-tools uses it in the `static_host` conf.py setting. Custom build scripts could also use the environment variable.
https://github.com/apluslms/gitmanager/blob/912a7fceff8b232445bbe6cb6d5791f2f4daccf3/builder/builder.py#L112
The value of STATIC_CONTENT_HOST is like:
`https://gitmanager.cs.aalto.fi/static/<coursekey>`
where `<coursekey>` is the correct course key.

PR #122, opened 2021-11-26
Remove gitmanager, add JWT login and external configuring WIP

### LINKED ISSUE #65: Implement API for course updates
>**Migrated from Trello**

API authentication should be done with JWT tokens.

Thus, mooc-grader should include rsa public key and then validate the JWT with it.

JWT token could include the course folder (e.g. `def_course` or `o1_2019`) in the `sub` or `course_id` field. Course identifier pair `(course_key, instance_key)` could be converted to a single word by replacing `_` with `__` and joining with single `_`. Result would be e.g. `course__key_instance__key` for above pair. For typical case of `('cs-a1110', 's2019')` it would be `cs-a1110_s2019`.

Currently, mooc-grader includes everything about the course: the static html files and exercise folders and such. With the API, we can presume there is no static files and we should focus only on the exercise data. Sadly, exercises in the build folders contain yaml files under `/_build/yaml` while the actual files are often under `/exercises`, but the real location is inside the `/_build/yaml` files.

The goal is to upload only the exercise data, thus everything under the `container.mount`. E.g. exercise directory `exercises/hello_python` for exercise defined in  `_build/yaml/m02_programming_exercises_02_hello_world_python.yaml` in `apluslms/course-templates` course repository.

Later on, we should do the similar task for uploading the static files to a static file hosting service (to be implemented).

Thus, `index.yaml` would be read by at least these deployment containers:

* mooc-grader to upload course exercises used by grader
* static-host to upload course static data
* aplus to upload configuration to A+, requires full URLs from previous steps

As all of these steps read and validate `index.yaml`, it would be good idea to implement some generic tool with separate implementation for specific backends.

Tools to consider:

* https://www.django-rest-framework.org/
* https://github.com/GetBlimp/django-rest-framework-jwt (or similar)
* Using swagger/openapi to document the API:
  * https://swagger.io/ and https://www.openapis.org/
  * https://django-rest-swagger.readthedocs.io/en/latest/

An example implementation in context of updating LTI keys:

* code: https://github.com/Aalto-LeTech/django-lti-login/blob/jwt-api/django_lti_login/api_views.py
* docs: https://github.com/Aalto-LeTech/django-lti-login/tree/jwt-api#a-lti-key-management-api

----
**Comments**

>**Jaakko Kantojärvi :**
For testing, this script can be used to generate RSA key, public key and JWT token for it. e.g. ./create_jwt.sh def_course@shepherd would create key with payload:
```json
{
  "sub": "def_course",
  "iss": "shepherd",
  "jwi": "dc4eb202-634b-4c7b-985f-4ad47554d95b",
  "iat": 1562070739,
  "exp": 1564662739
}
```
> Note that, in this use case, we might not be able to use expire timestamp.
> https://gist.github.com/raphendyr/15d40bb30960edfca2eb6238031ee5fb
>* Use target branchcloud-dev when ever creating PR for the card.

lainets:
# Description

**What?**

Removes gitmanager, and adds JWT authentication/authorization and configuring of courses through http.

**Why?**

Gitmanager has been made into a separate service. This also requires configuring of the exercises through http. JWT login is added so that not anyone can do that.

**How?**

JWT login and permission management through aplus-auth library.

Fixes #65. These should be moved to gitmanager repo: #63, #64, #57, #31, #27. Probably #66 as well.

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

I tested that the exercises and http configuring work.

- [X] Chrome
- [ ] Firefox

**Think of what is affected by these changes and could become broken**

# Translation

- [] Did you modify or add new strings in the user interface? ([Read about how to create translation](https://github.com/apluslms/a-plus/tree/master/doc#running-tests-and-updating-translations))

It was decided that no translation for the error messages is neccesary.

# Programming style

- [X] README.md.
- [ ] Aplus Manual.

WIP

# Is it Done?


lainets:
The external course configuration uses a bit different exercise loader. Now that I think about it, if some course needs their own exercise loader, this will not work. This is also related to #60 in that settings the exercise loader shouldn't probably be allowed anyway.

lainets:
Notable is that, for externally configured courses, the course keys are replaced by the A+ DB id. This means URL's have a number instead of a readable key. Is this a problem? If it is, we could add symlinks to circumvent this.

lainets:
Git webhook was removed with the gitmanager. The id affects all URLs on courses that are configured externally. In that case, the static files, chapter HTML files and aplus-json are provided by some other service (gitmanager). The only relevant one is the exercise service URL, which has the numeric id instead of a readable course key

access/views.py
```diff
@@ -276,11 +400,24 @@ def children_recursion(parent):
...
+        response = render(request, 'access/login.html')
+        response.delete_cookie("AuthToken")
+        return response
+
+    def post(self, request):
+        if not hasattr(request, "user") or not request.user.is_authenticated:
+            return HttpResponse("Invalid token", status=401)
+        else:
+            response = HttpResponse()
+            response.set_cookie("AuthToken", str(request.auth))
```
Mankro, line 417: Should the cookie security be improved with parameters `secure=True, httponly=True`? What about expiration time?

https://docs.djangoproject.com/en/3.2/ref/request-response/#django.http.HttpResponse.set_cookie

access/views.py
```diff
@@ -39,10 +46,125 @@ def index(request):
...
-        "manager": 'gitmanager' in settings.INSTALLED_APPS,
     })
 
 
+def publish(request):
+    """
+    Move a course from store to the main folder
+    """
+    if request.method != "POST":
+        raise Http404()
```
Mankro, line 57: HTTP 405 (Method Not Allowed) would be more appropriate, right?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/405

access/views.py
```diff
@@ -39,10 +46,125 @@ def index(request):
...
+
+    if "course_id" not in request.POST:
+        return HttpResponse("Missing course_id", status=400)
+
+    course_id = request.POST["course_id"]
+
+    if not request.auth.permissions.instances.has(Permission.WRITE, id=int(course_id)):
```
Mankro, line 64: If `course_id == "qwerty"`, this will crash. HTTP POST parameters could be any strings. You should validate the type first.

access/views.py
```diff
@@ -39,10 +46,125 @@ def index(request):
...
+    store_root_dir = Path(settings.COURSE_STORE)
+    course_path = root_dir / course_id
+    store_course_path = store_root_dir / course_id
+    version_id_path = root_dir / (course_id + ".version")
+    store_version_id_path = store_root_dir / (course_id + ".version")
+
+    store_version_id = None
+    if store_version_id_path.exists():
+        with open(store_version_id_path) as f:
+            store_version_id = f.read()
```
Mankro, line 77:
```suggestion
    try:
        with open(store_version_id_path) as f:
            store_version_id = f.read()
    except OSError:
        # how should the error be handled?
        return HttpResponse("Failed")
```
File IO in this case should absolutely use try-except instead of assuming that the file can not be removed between `exists()` and `read()`.

access/views.py
```diff
@@ -39,10 +46,125 @@ def index(request):
         })
     return render(request, 'access/ready.html', {
         "courses": courses,
-        "manager": 'gitmanager' in settings.INSTALLED_APPS,
     })
 
 
+def publish(request):
```
Mankro, line 52: Is `login_required` missing on purpose?

access/views.py
```diff
@@ -39,10 +46,125 @@ def index(request):
...
+        renames([
+            (store_version_id_path, version_id_path),
+            (store_course_path, course_path),
+        ])
+
+        return HttpResponse()
+
+    version_id = None
+    if version_id_path.exists():
+        with open(version_id_path) as f:
```
Mankro, line 89: Must handle OSError here.

access/views.py
```diff
@@ -39,10 +46,125 @@ def index(request):
...
+            version_id = f.read()
+
+    if version_id == request.POST.get("version_id"):
+        return HttpResponse()
+    else:
+        return HttpResponse("Unknown version id", status=400)
```
Mankro, line 95: Would status 404 make sense?

access/views.py
```diff
@@ -39,10 +46,125 @@ def index(request):
...
+
+
+@login_required
+def configure(request):
+    '''
+    Configure a course according to the gitmanager protocol.
+    '''
+    if request.method != "POST":
+        raise Http404()
```
Mankro, line 104: Same as above, 405 Method Not Allowed.

access/views.py
```diff
@@ -39,10 +46,125 @@ def index(request):
...
+
+    if "publish" in request.POST and request.POST["publish"]:
```
Mankro, line 106:
```suggestion
    if request.POST.get("publish"):
```

access/views.py
```diff
@@ -39,10 +46,125 @@ def index(request):
...
+        return publish(request)
+
+    if "exercises" not in request.POST or "course_id" not in request.POST:
+        return HttpResponse("Missing exercises or course_id", status=400)
+
+    exercises = json.loads(request.POST["exercises"])
```
Mankro, line 112: Must handle JSON parsing errors. Don't crash the server if (user) input is invalid.

access/views.py
```diff
@@ -39,10 +46,125 @@ def index(request):
...
+    course_id = request.POST["course_id"]
+
+    if not request.auth.permissions.instances.has(Permission.WRITE, id=int(course_id)):
```
Mankro, line 115: Validate that `course_id` is an integer or else we crash.

access/views.py
```diff
@@ -39,10 +46,125 @@ def index(request):
...
+        return HttpResponse(status=401)
+
+    root_dir = Path(settings.COURSE_STORE)
+    course_path = root_dir / course_id
+    if course_path.exists():
+        rmtree(course_path)
```
Mankro, line 121:
```suggestion
    try:
        rmtree(course_path)
    except OSError:
        logger.exception(...)
        # can we continue after this or should we return from the function?
```
`shutil.rmtree` has the parameter `ignore_errors` that might be useful in this case to avoid Exceptions.

access/views.py
```diff
@@ -39,10 +46,125 @@ def index(request):
...
+        tarh = TarFile(fileobj=tar_file)
+        tarh.extractall(course_files_path)
+
+    course_config = {
+        "name": course_id,
+        "exercises": [ex["key"] for ex in exercises],
+        "exercise_loader": "access.config._ext_exercise_loader",
+    }
+
+    with open(course_path / "index.json", "w") as f:
```
Mankro, line 140: Must handle OSError. And in the IO operations below.

access/views.py
```diff
@@ -39,10 +46,125 @@ def index(request):
...
+    course_config = config._course_root_from_root_dir(course_id, root_dir)
+
+    defaults = {}
+    for info in exercises:
+        of = info["spec"]
+        if info.get("config"):
+            of["config"] = info["key"] + ".json"
+            course, exercise = config.exercise_entry(course_config, info["key"], "_root")
+            of = export.exercise(request, course, exercise, of)
+        defaults[of["key"]] = of
```
Mankro, line 162: It wasn't clear to me why all the configs are exported to the JsonResponse. Does the new Git manager use them? (Does the Git manager call this configure view?)

Mankro (changes requested):
Great work! I had many comments, but they are smallish issues.

* File IO error handling. You have often ignored error handling completely in file operations, but I think they should be properly handled. I don't want the server to crash completely if reading or writing a file fails sometimes. The system should try to clean up any corrupt/invalid files, **log errors** and respond with some error message/status to the http client. We are supposed to avoid crashing servers and responding with the status 500 (which results from a Python crash) when something goes wrong. Logging is important in error cases!
* Have you used the run-mooc-grader Docker container? It will require some changes in the settings since this PR changes settings.py.
* commit messages should start with a capital letter, but no need to change this afterwards in this PR. Example message: `Remove gitmanager` (good: capital R, imperative verb, no ending period `.`)
* One commit message caught my attention. Commit `add utility functions for file handling`: `quarantee` -> `guarantee` (don't make us quarantined, guarantee vs quarantine :) (though, what does "afterwards guarantee" even mean?)

util/files.py
```diff
@@ -160,3 +154,80 @@ def read_and_remove_submission_meta(sid):
...
+
+    src, dst = os.fspath(src), os.fspath(dst)
+    if not os.path.exists(dst) or (os.path.isfile(dst) and os.path.isfile(src)):
+        if keep_tmp and os.path.exists(dst):
+            tmpdst = _tmp_path(dst)
+            os.rename(dst, tmpdst)
+
+        try:
+            os.rename(src, dst)
+        except:
```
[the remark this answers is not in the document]
lainets: In this case, I wanted to catch _everything_ but I guess there really shouldn't be any other exceptions that OSError

util/files.py
```diff
@@ -160,3 +154,80 @@ def read_and_remove_submission_meta(sid):
...
+            if tmpdst:
+                os.rename(tmpdst, dst)
```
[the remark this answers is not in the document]
lainets: Then the exception is bubbled up? We will raise an exception anyway, and there is nothing to do if the rename in the except fails

util/files.py
```diff
@@ -18,28 +22,76 @@ def random_ascii(length, rng=None):
...
-    d = datetime.datetime.now().strftime("%Y%m%d%H%M%S%f") + random_ascii(5)
-    submission_dir = os.path.join(settings.SUBMISSION_PATH,
-        course["key"], exercise["key"], d)
-
-    # Create empty directory.
-    if not os.path.exists(submission_dir):
-        os.makedirs(submission_dir)
-
-    return submission_dir
+class SubmissionDir:
```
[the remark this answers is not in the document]
lainets: The SubmissionDir class is effectively just a copy of the original code but wrapped inside a class. I don't think it is a good idea to spend time on those now. My additions, on the other hand, either catch exceptions or deliberately let them bubble up, so that the calling code knows what went wrong.

util/files.py
```diff
@@ -160,3 +154,80 @@ def read_and_remove_submission_meta(sid):
...
+
+def rename(src: PathLike, dst: PathLike, keep_tmp=False) -> Optional[str]:
+    """
+    renames a file or directory while making sure that the destination will only be removed if successful.
+    returns None if keep_tmp = False. Otherwise, returns a tmp path to dst (None if dst does not exist).
+    """
+    tmpdst = None
+
+    src, dst = os.fspath(src), os.fspath(dst)
+    if not os.path.exists(dst) or (os.path.isfile(dst) and os.path.isfile(src)):
```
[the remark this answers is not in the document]
lainets: It is technically possible for that to happen but the probability is very small. There is also not much we can do if something else removes the file. We could make this function 100 lines long with a dozen different checks but I don't think it is worth it.

scripts/run-template.py
```diff
@@ -0,0 +1,37 @@
...
+    out and err as in stdout and stderr output of a program.
+
+    **kwargs is recommended for the case that more arguments are added later.
+    """
+    try:
+        # implement here
+        ...
+    except:
+        logger.exception("An exception while trying to run grading container")
+        return 1, "", "An exception happened"
```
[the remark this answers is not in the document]
lainets: It doesn't really matter as the error is just logged anyway but I'll add it.

scripts/docker-run.py
```diff
@@ -0,0 +1,81 @@
...
+            remove = True,
+            detach = True,
+            environment = {
+                "SID": submission_id,
+                "REC": host_url,
+            },
+            volumes = volumes,
+        )
+
+        return 0, "", ""
```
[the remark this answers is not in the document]
lainets: The documentation for docker_client.containers.run says that there is no output for detach = True. I believe it just raises an error if there is a problem

access/views.py
lainets: Yes, this is called from configure which has login_required instead of being directly accessible.

access/views.py
lainets: Yes, gitmanager calls this. The response is the default (not specified in course configuration) values for the exercise configs passed to A+.

util/login_required.py
```diff
@@ -0,0 +1,7 @@
+from typing import Callable
+from aplus_auth.auth.django import login_required as login_required_base
+from django.http.response import HttpResponse
+
+
+ViewType = Callable[..., HttpResponse]
+login_required: Callable[[ViewType],ViewType] = login_required_base(redirect_url="/login?referer={url}")
```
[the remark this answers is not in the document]
lainets: Postpone for some other time

access/types/stdasync.py
```diff
@@ -296,51 +303,62 @@ def _acceptSubmission(request, course, exercise, post_url, sdir):
...
-        "key": course["key"],
-        "name": course["name"],
-    }
-    exercise_extra = {
-        "key": exercise["key"],
-        "title": exercise.get("title", None),
-        "container_config": c
+    ro_mounts = c.get("mounts", {}).copy()
+    if not isinstance(ro_mounts, dict):
+        raise ConfigError("'container'->'mounts' must not be specified or be a <path on course>-<mount path> dictionary")
```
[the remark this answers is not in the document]
lainets: Better now?

access/templates/access/login.html
```diff
@@ -0,0 +1,38 @@
+{% extends 'grader_frame.html' %}
+{% load i18n %}
+
+{% block body %}
+<div class="page-header">
+	<h1>Login</h1>
+</div>
+<form id="login-form">
+    Access token: <input type="text" name="token"></input>
```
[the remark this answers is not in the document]
Mankro: Yes, skip this.

access/types/stdasync.py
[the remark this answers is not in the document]
Mankro: Good, thanks!

Mankro (approved):
Thanks! I will merge this now. More fixes can be done later, but this is already in a good state.

util/login_required.py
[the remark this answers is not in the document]
lainets: I fixed the escaping on aplus-auth but reversing the path doesn't seem to work. I'm guessing the problem is that the 'login' view hasn't been defined yet when this module is first loaded. I'd say this is such a minor issue that we can ignore it

access/views.py
lainets: I didn't include expiration time because the token itself has one. The others I added

util/login_required.py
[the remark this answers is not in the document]
Mankro: OK, ignore for now!

PR #125, opened 2022-01-07
Bug fixes

lainets:
# Description

**What?**

Couple bug fixes. 

**Why?**

There were problems with finding templates and importing modules.

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

- [ ] Chrome
- [ ] Firefox
- [X] This pull request cannot be tested in the browser.

# Programming style

# Is it Done?

### DIFF
--- a/grader/settings.py
+++ b/grader/settings.py
@@ -72,7 +72,7 @@ TEMPLATES = [
         'DIRS': [
             join(BASE_DIR, 'local_templates'),
             join(BASE_DIR, 'templates'),
-            join(BASE_DIR, 'courses'),
+            # COURSES_PATH is added at the bottom of this file
         ],
         'APP_DIRS': True,
         'OPTIONS': {
@@ -242,6 +242,10 @@ assert AJAX_KEY, "Secure random string is required in AJAX_KEY"
 APLUS_AUTH.update(APLUS_AUTH_LOCAL)
 
 Path(COURSE_STORE).mkdir(parents=True, exist_ok=True)
+
+# Add COURSES_PATH to template dirs here because at this point, the local settings have been imported.
+if len(TEMPLATES) == 1 and 'DIRS' in TEMPLATES[0]:
+    TEMPLATES[0]['DIRS'].append(COURSES_PATH)
 
 # Drop x-frame policy when debugging
 if DEBUG:
--- a/grader/settings.py
+++ b/grader/settings.py
@@ -241,7 +241,13 @@ assert AJAX_KEY, "Secure random string is required in AJAX_KEY"
 
 APLUS_AUTH.update(APLUS_AUTH_LOCAL)
 
-Path(COURSE_STORE).mkdir(parents=True, exist_ok=True)
+try:
+    Path(COURSE_STORE).mkdir(parents=True, exist_ok=True)
+except OSError as e:
+    raise Exception(
+      "Cannot create COURSE_STORE directory (is the filesystem read-only?)."
+      "Create it manually or fix the issue"
+    ) from e
 
 # Add COURSES_PATH to template dirs here because at this point, the local settings have been imported.
 if len(TEMPLATES) == 1 and 'DIRS' in TEMPLATES[0]:
--- a/util/importer.py
+++ b/util/importer.py
@@ -8,7 +8,20 @@ import sys
 from django.conf import settings
 
 
-def import_path(path: str, module_name: str = "", force_reload = False):
+# This is needed so that importing course files (for custom views) works
+# correctly. We actually can import and run a file without this, as it is
+# importing the file directly through its file location. Problems arise when
+# the imported module itself tries to import something from the course
+# directory. E.g. if example_course/exercises/iframe.py tries to do
+# 'from .test import MyClass' or 'from views.test import MyClass', we'd get
+# an error. My understanding is that this happens because it cannot find the
+# example_course folder in sys.path, so it cannot import it (and therefore its
+# descendants).
+if settings.COURSES_PATH not in sys.path:
+    sys.path.append(settings.COURSES_PATH)
+
+
+def import_path(path: str, module_name: str = "", force_reload: bool = False):
     """Imports a module from a specified path"""
     if not force_reload and module_name and module_name in sys.modules:
         return sys.modules[module_name]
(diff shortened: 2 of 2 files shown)


util/importer.py
```diff
@@ -8,7 +8,13 @@
 from django.conf import settings
 
 
-def import_path(path: str, module_name: str = "", force_reload = False):
+# this is needed so that import from course files works correctly.
+# it seems like importing a module requires the whole dotted path to a
+# path in sys.path, otherwise imports within the module's package fail
+sys.path.append(settings.COURSES_PATH)
```
Mankro, line 14: Could you give an example of what kind imports this is about?

For example, was this about JSAV exercises that define a custom view type (Python function in the course repo)? MOOC-Grader exercise view would then import the Python module in the course repo.

Is there a reason why `sys.path` is modified in this module specifically? Could this be done in settings.py?

grader/settings.py
```diff
@@ -252,7 +252,17 @@
...
-Path(COURSE_STORE).mkdir(parents=True, exist_ok=True)
+try:
+    Path(COURSE_STORE).mkdir(parents=True, exist_ok=True)
+except OSError as e:
+    raise Exception(
+      "Cannot create COURSE_STORE directory (is the filesystem read-only?)."
+      "Create it manually or fix the issue"
+    ) from e
+
+# add COURSES_PATH to template dirs here because we only now have read the local settings
```
Mankro, line 263:
```suggestion
# Add COURSES_PATH to template dirs here because at this point, the local settings have been imported.
```

util/importer.py
Mankro: Why did the imports work before v1.12 without this sys.path change?

Mankro (commented):
I have questions below before merging. The sys.path change is confusing; why is it needed now?

util/importer.py
lainets: It's about any custom view import. E.g. the JSAV exercises. They worked before because it was assumed that COURSES_PATH is courses/ but it isn't anymore. The modification should be possible in settings.py as well but I put it here as it pertains to the importing functionality that this file is about. E.g. when reading the source for the import functions, I think it is useful to have this nearby.

Mankro: OK. Is it guaranteed that `sys.path.append()` is executed only once even when the module `util/importer.py` is imported in many other modules?

lainets: This is guaranteed, at least normally. It is possible that Django, uWSGI or some other framework does something unexpected. I added an if check just in case.

util/importer.py
```diff
@@ -8,7 +8,14 @@
 from django.conf import settings
 
 
-def import_path(path: str, module_name: str = "", force_reload = False):
+# this is needed so that import from course files works correctly.
+# it seems like importing a module requires the whole dotted path to a
+# path in sys.path, otherwise imports within the module's package fail
```
Mankro, line 13:
```suggestion
# This is needed for importing modules defined in the course repo.
# For example, the exercise `view_type` may refer to Python modules in the course repo.
# The path to course-specific view types is given relative to the course directory.
# For example, `view_type: .exercises.jsav.views.iframe.view`
```
I think the comment was not very easy to understand. Is my version correct?

By the way, if the Python code runs `import exercises.jsav.views.iframe`, then doesn't it make a lot of sense that PYTHONPATH must include the course root directory (because the exercises subdirectory starts from there)? If PYTHONPATH contained the jsav directory, then it would NOT find package `exercises` there.

The `import_named` function actually adds the course key (directory) to the path before importing, thus the import statement to run is like `import traky.exercises.jsav.views.iframe`. Therefore, PYTHONPATH must contain `settings.COURSES_PATH`.

But does the MOOC-Grader now contain subdirectories `publish` and `store` under `COURSES_PATH` so that the course directories are not directly under `COURSES_PATH`?

lainets: So, this is a bit complicated. We actually can import and run the exercises.jsav.views.iframe file using the code that we already have, as it is importing the file directly through its file location and not doing 'import exercises.jsav.views.iframe'. The problems arise when the iframe module itself tries to import something from the course directory. E.g. if iframe.py tries to do 'from .test import MyClass', we would get an error 'traky is not a module', or if it tries to do 'from views.test import MyClass', we'd get 'views is not a package'. My understanding is that this happens because it cannot find the traky folder in sys.path, so it cannot import it (and therefore its descendants).

> By the way, if the Python code runs import exercises.jsav.views.iframe, then doesn't it make a lot of sense that PYTHONPATH must include the course root directory (because the exercises subdirectory starts from there)? If PYTHONPATH contained the jsav directory, then it would NOT find package exercises there.

See above for comment about ' import exercises.jsav.views.iframe'.

Yes, it wouldn't find exercises package if the directory above it isn't in sys.path or the package above exercises hasn't been loaded yet. We shouldn't need the course directory itself in sys.path as we are including it in the import path when we load the view (e.g. traky.exercises.jsav.views.iframe)

> The import_named function actually adds the course key (directory) to the path before importing, thus the import statement to run is like import traky.exercises.jsav.views.iframe. Therefore, PYTHONPATH must contain settings.COURSES_PATH.

See above for the explanation as to why COURSES_PATH is need in sys.path.

> But does the MOOC-Grader now contain subdirectories publish and store under COURSES_PATH so that the course directories are not directly under COURSES_PATH?

COURSES_PATH is the publish/ directory.

Mankro: Thanks for the explanation!

> COURSES_PATH is the publish/ directory.

Not in the default `grader/settings.py`, though? Shouldn't that be fixed?

You have force-pushed a couple of times, but your branch actually isn't rebased onto the upstream master. You could that now.

lainets: In the default settings courses/ is the publish directory, and course_store/ is the store directory. The publish directory is courses/publish/ only in the Aalto production server

Mankro: > In the default settings courses/ is the publish directory, and course_store/ is the store directory. The publish directory is courses/publish/ only in the Aalto production server

OK. Why was it done this way? Network file system?

lainets: The publish and store directories need to be on the same filesystem for the os.rename to work. We thus need the store and publish directories to be under the same directory. Each directory under the project root is mounted separately to allow separation of config from other files, so they are both under the courses directory.

PR #126, opened 2022-01-12
Update aplus-auth to 0.2.0, add permission checks to views, add logging, fix custom template caching

lainets:
# Description

**What?**

Update aplus-auth to 0.2.0, add permission checks to views, add logging, and fix custom template caching.

**Why?**

Shortens JWT lengths, and increases security.
Security logging was required.
Templates wouldn't be reloaded after they were changed.

**How?**

Respond with 403 if the course key is an integer (i.e. it is the remote id) and the user has no access.
Logging on endpoints where there could be an effect to the course.
Remove caching of courses' custom templates.

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

- [ ] Chrome
- [ ] Firefox
- [X] This pull request cannot be tested in the browser.

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/access/urls.py
+++ b/access/urls.py
@@ -1,19 +1,33 @@
-from django.conf.urls import url
+from django.urls import path, register_converter
 
 from access import views
+from access.converters import BasenameConverter
+
+register_converter(BasenameConverter, "basename")
 
 urlpatterns = [
-    url(r'^$', views.index, name='index'),
-    url(r'^configure$', views.configure, name='configure'),
-    url(r'^test-result$', views.test_result, name='test-result'),
-    url(r'^container-post$', views.container_post, name='container-post'),
-    url(r'^ajax/([\w-]+)/([\w-]+)$', views.exercise_ajax, name='ajax'),
-    url(r'^model/([\w-]+)/([\w-]+)/([\w\d\_\-\.]*)$', views.exercise_model, name='model'),
-    url(r'^exercise_template/([\w-]+)/([\w-]+)/([\w\d\_\-\.]*)$', views.exercise_template, name='exercise_template'),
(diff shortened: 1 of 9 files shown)


access/urls.py
```diff
@@ -1,19 +1,33 @@
-from django.conf.urls import url
+from django.urls import path, register_converter
 
 from access import views
+from access.converters import BasenameConverter
+
+register_converter(BasenameConverter, "basename")
```
Mankro, line 6: Nice work with the URLs here!

access/views.py
```diff
@@ -59,10 +68,13 @@ def publish(request):
         return HttpResponse(status=405)
 
     if "course_id" not in request.POST:
+        SecurityLog.info(request, f"PUBLISH")
```
Mankro, line 71: Does the log message describe if the request was accepted or declined?

access/views.py
```diff
@@ -481,6 +507,7 @@ def test_result(request):
     return HttpResponse(result or 'No test result received yet.')
 
 
+@instance_read_access_required
```
Mankro, line 510: The generated exercise files are supposed to be public to the web so that the students may download them. The download link in the A+ page points directly to this mooc-grader endpoint. So the login decorator should be removed, right?

As an example, see the generated file links inside these personalized exercises:
https://plus.cs.aalto.fi/aplus-manual/master/programming_exercises/personalized_exercises/

Mankro (changes requested):
Great work! I have a question about the security log and I think the generated file view can't require any authentication.

access/views.py
lainets: Good now? Sorry about the rebase, the changes were pretty big. Only the views.py and log.py files changed.

Mankro (approved):
Thanks!

PR #139, opened 2022-11-29
Replace harcoded login urls

### LINKED ISSUE #136: Hardcoded /login URLs do not work with FORCE_SCRIPT_NAME
When the mooc-grader installation uses the `FORCE_SCRIPT_NAME` Django setting to serve all mooc-grader views behind a common URL prefix, then the system breaks down due to the hardcoded `/login` URL in a couple of places. The code should use the Django URL resolvers to form the URLs instead of hardcoding the URL strings.

This affects Tampere University since they use `FORCE_SCRIPT_NAME`. Their graders are supposed to be served from URL paths starting `/graderA`, on another server `/graderE` and so on. The prefix should affect all URLs of the MOOC-Grader. The URL path prefix is needed when the URL domain name is shared between multiple servers.

Hardcoded `/login` is used here:

https://github.com/apluslms/mooc-grader/blob/1f5b16365930f5b18de4c39e427ce2be7e0884d6/util/auth.py#L11
When Tampere hardcodes `login_required_base(redirect_url="/graderA/login?referer={url}")` there, it works.
Note that `reverse()` doesn't seem to work in the module-level code since Django has not set up everything when the module is loaded and the module-level code is executed.
https://docs.djangoproject.com/en/3.2/ref/urlresolvers/#reverse

https://github.com/apluslms/mooc-grader/blob/1f5b16365930f5b18de4c39e427ce2be7e0884d6/access/templates/access/login.html#L20

Aalto internal ticket:
https://rt.cs.aalto.fi/Ticket/Display.html?id=21827

EerikSaksi:
Rather than hardcoding /login/, use the url template tag and reverse_lazy

Fixes #136 

# Description

In a few locations, the hardcoded /login/ url was used.

**Why?**

If the login url used is different from /login/ this causes issues

**How?**

Using the `{% url %}` tag in the template, and `reverse_lazy` in the Python code. `reverse_lazy` is necessary as the url conf has not yet been loaded.

"It is useful for when you need to use a URL reversal before your project’s URLConf is loaded."
https://docs.djangoproject.com/en/4.1/ref/urlresolvers/#reverse-lazy

Fixes #136 

# Testing

**Remember to add or update unit tests for new features and changes.**

* How to [test your changes in A-plus](https://github.com/apluslms/a-plus/tree/master/doc#running-tests-and-updating-translations)
* How to [test accessibility](https://wiki.aalto.fi/display/EDIT/How+to+check+the+accessibility+of+pull+requests)

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

Ensuring that the template links to the same /login url despite changes, ensuring that the reverse_lazy does not cause issues.

- [X] Chrome
- [ ] Firefox

**Think of what is affected by these changes and could become broken**

# Translation

- [ ] Did you modify or add new strings in the user interface? ([Read about how to create translation](https://github.com/apluslms/a-plus/tree/master/doc#running-tests-and-updating-translations))

# Programming style

- [X] Did you use Python type hinting in all functions that you added or edited? ([type hints](https://docs.python.org/3/library/typing.html) for function parameters and return values)

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/access/templates/access/login.html
+++ b/access/templates/access/login.html
@@ -17,7 +17,7 @@
         const data = new FormData(e.target);
 
         const xml = new XMLHttpRequest();
-        xml.open("POST", "/login");
+        xml.open("POST", "{% url 'login' %}");
         xml.setRequestHeader("Authorization", "Bearer " + data.get("token"))
         xml.onload = () => {
             console.log(xml);
--- a/util/auth.py
+++ b/util/auth.py
@@ -6,9 +6,13 @@ from aplus_auth.auth.django import login_required as login_required_base
 from aplus_auth.payload import Permission
 from django.core.exceptions import PermissionDenied
 from django.http.response import HttpResponse
+from django.urls import reverse_lazy
+from django.utils.text import format_lazy
 
 ViewType = Callable[..., HttpResponse]
-login_required: Callable[[ViewType],ViewType] = login_required_base(redirect_url="/login?referer={url}")
+login_required: Callable[[ViewType], ViewType] = login_required_base(
+    redirect_url=format_lazy('{}?{}', reverse_lazy('login'), 'referer={url}'),
+)
 
 
 def access_check(request, permission: Permission, course_key: Union[str,int]):


access/templates/access/login.html
```diff
@@ -17,7 +17,7 @@ <h1>Login</h1>
         const data = new FormData(e.target);
 
         const xml = new XMLHttpRequest();
-        xml.open("POST", "/login");
+        xml.open("POST", {% url 'login' %});
```
Mankro, line 20: No quotes around the URL even though this is JavaScript code?

The output JavaScript looks like this: `xml.open("POST", /login);`
As you can see, the string literal `/login` is missing the quotes `" "`. This causes a syntax error and the JS crashes.

I think this should work like this:
`xml.open("POST", "{% url 'login' %}");`
I just added quotes around the output of the `url` template tag.

util/auth.py
```diff
@@ -6,9 +6,11 @@
 from aplus_auth.payload import Permission
 from django.core.exceptions import PermissionDenied
 from django.http.response import HttpResponse
+from django.urls import reverse_lazy
 
 ViewType = Callable[..., HttpResponse]
-login_required: Callable[[ViewType],ViewType] = login_required_base(redirect_url="/login?referer={url}")
+login_required: Callable[[ViewType], ViewType] = login_required_base(
+    redirect_url=reverse_lazy('login', kwargs={'referer': '{url}'}))
```
Mankro, line 13: Does this really work for you? I get a crash on the MOOC-Grader front page when I haven't previously logged into the grader:

```
NoReverseMatch at /

Reverse for 'login' with keyword arguments '{'referer': '{url}'}' not found. 1 pattern(s) tried: ['login\\Z']
```

The URL query parameters (after the `?` question mark) are not part of the Django URL resolver. The parameters need to be appended to the output of the `reverse()` function. Since `reverse_lazy` does not return a normal string due to the lazy evaluation, then maybe you need to create a new lazy string that appends the URL query parameters to the output after `reverse_lazy()`'s lazy string has been evaluated.

Read these too:
* https://stackoverflow.com/questions/4995279/including-a-querystring-in-a-django-core-urlresolvers-reverse-call
* https://stackoverflow.com/questions/29252312/how-can-i-join-lazy-translation-in-django

Does this work? The StackOverflow link above suggests using `format_lazy()` for concatenating lazy strings.
```python
login_required: Callable[[ViewType], ViewType] = login_required_base(
    redirect_url=format_lazy('{}?{}', reverse_lazy('login'), 'referer={url}'),
)
```
https://docs.djangoproject.com/en/3.2/ref/utils/#django.utils.text.format_lazy

I'm not sure if the `login_required_base()` function works correctly with the lazy parameter `redirect_url`. Try it out!
https://github.com/apluslms/aplus-auth/blob/8050cecd99dd9335adf684a2ec57c068e7a67edd/src/aplus_auth/auth/django.py#L41

Mankro (changes requested):
Brilliant, I hadn't realized there is a function `reverse_lazy()` for these situations! But your implementation does not yet work correctly. See the comments below.

util/auth.py
EerikSaksi: Login required_base seems to work with the lazy formatting

Mankro (approved):
Great!

PR #144, opened 2023-06-05
Remove and clean up old queue and wait variables

lainets:
They seem to be relics from a time the system used celery instead of docker.

The change basically replaces the `wait` and `queue < 3` statements with True, and omits any unreachable code given that.

I only did very basic test of checking that I can still submit things because of how simple the change is.

### DIFF
--- a/access/templates/access/async_accepted.html
+++ b/access/templates/access/async_accepted.html
@@ -2,17 +2,7 @@
 {% load i18n %}
 
 {% block exercise %}
-{% if result.synchronous %}
-<div class="alert alert-danger">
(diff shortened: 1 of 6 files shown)


Mankro (approved):
Nice!

Celery was used with the chroot sandbox. The chroot sandbox was removed in https://github.com/apluslms/mooc-grader/pull/58.

courses/README.md still contains a reference to the chroot Celery queue that could be cleaned up:
https://github.com/apluslms/mooc-grader/blob/e6fdfe826acd50b055148a696587afad619cb53c/courses/README.md?plain=1#L188

PR #151, opened 2023-08-09
Improve generic feedback on pick-any questions

### LINKED ISSUE #150: Improve generic feedback on pick-any questions
This issue description is copied from [a previous a-plus-rst-tools issue](https://github.com/apluslms/a-plus-rst-tools/issues/157), since the fix is easier to implement here:

In pick-any questions (i.e., multiple checkboxes), it is often—always?—a good idea to give points only after the student has answered fully correctly (i.e., ticked all the right boxes), and to give zero points otherwise. However, students do sometimes assume that zero points means “nothing’s correct”; moreover, they sometimes expect to get partial credit.

IMO, it would be a good idea to insert an additional feedback bullet on pick-anys, to be displayed whenever a student’s answer is partially correct (i.e., something was right, but not everything). Maybe something like this: “Fully correct answers are awarded with a small number of points; the question is meant as a whole, so no partial points are given. A zero score does not mean that nothing in your answer was correct.”

ihalaij1:
# Description

**What?**

Add an additional feedback note on pick-anys when partial-points option is not used and the answer is incorrect.

image

**Why?**

Students sometimes assume that zero points means “nothing is correct”; moreover, they sometimes expect to get partial credit.

Fixes #150

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [x] Manual testing.

Tested that the new feedback note only shows in pick-any questions without partial-points option when the answer is incorrect.

- [x] Chrome
- [ ] Firefox

# Translation

- [x] Did you modify or add new strings in the user interface? ([Read about how to create translation](https://github.com/apluslms/a-plus/tree/master/doc#running-tests-and-updating-translations))

# Programming style

- [x] Did you follow our [style guides](https://apluslms.github.io/contribute/styleguides/)?
- [ ] Did you use Python type hinting in all functions that you added or edited? ([type hints](https://docs.python.org/3/library/typing.html) for function parameters and return values)

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/access/templatetags/access.py
+++ b/access/templatetags/access.py
@@ -35,4 +35,7 @@ def find_common_hints(list_of_hints, checked):
                     hints_seen.add(hint_value)
                     hints.append(hint_text)
 
+        if list_of_hints.get('points_only_for_fully_correct'):
+            hints.append(list_of_hints.get('points_only_for_fully_correct'))
+
     return hints
--- a/access/types/forms.py
+++ b/access/types/forms.py
@@ -683,6 +683,14 @@ class GradedForm(forms.Form):
             else:
                 hints.append(_('HINT_MULTIPLE_CHOICES_SELECTABLE'))
 
+        if t == "checkbox" and not ok and not configuration.get("partial_points"):
+            # Show this hint in checkbox questions when the answer is not
+            # correct and the partial points option is not set.
+            if checkbox_feedback:
+                hints['points_only_for_fully_correct'] = _('POINTS_AWARDED_ONLY_FOR_FULLY_CORRECT_ANSWERS')
+            else:
+                hints.append(_('POINTS_AWARDED_ONLY_FOR_FULLY_CORRECT_ANSWERS'))
+
         if name in self.fields:
             self.fields[name].grade_points = earned_points
             self.fields[name].max_points = points
--- a/locale/en/LC_MESSAGES/django.po
+++ b/locale/en/LC_MESSAGES/django.po
@@ -261,7 +261,14 @@ msgstr ""
 
 #: access/types/forms.py
 msgid "HINT_MULTIPLE_CHOICES_SELECTABLE"
-msgstr "Multiple choices are selectable"
+msgstr "Multiple choices are selectable."
+
+#: access/types/forms.py
+msgid "POINTS_AWARDED_ONLY_FOR_FULLY_CORRECT_ANSWERS"
+msgstr ""
+"Only fully correct answers are awarded with points; the question is meant as "
+"a whole, so no partial points are given. A zero score does not necessarily "
+"mean that your answer was entirely incorrect."
 
 #: access/types/forms.py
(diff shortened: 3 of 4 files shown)


ihalaij1:
I noticed that the note admonition is not displayed as a blue box on the individual exercise page (the CSS class is missing there?).
Is this minor problem something that needs to be fixed?

Mankro:
> I noticed that the note admonition is not displayed as a blue box on the individual exercise page (the CSS class is missing there?). Is this minor problem something that needs to be fixed?

That is because the styles for the HTML class `admonition note` are only defined in the default.css file in A-plus-rst-tools:
https://github.com/apluslms/a-plus-rst-tools/blob/9894bb18ca9a940455a132275b5ce7366847582a/theme/aplus/static/default.css#L53-L59

The A-plus-rst-tools styles only affect chapter pages since they are rendered to HTML by the rst tools (Sphinx).

This problem can be fixed by using the existing questionnaire feedback styles instead of adding a new info admonition. I would prefer to use the existing styles anyway.

Mankro (changes requested):
I think the message could be added as an additional feedback item to the question. Now, you have added a separate implementation for a very specific message that is functionally similar to feedback items, but it is implemented and rendered completely differently than the other feedback items. In addition, I feel that the blue info box is too verbose and takes a lot of space on the screen. If it were rendered like other feedback items, it would blend in better and simply consume less screen space.

You don't need to change any HTML template since they already render the hints. You just add this new hint in the questionnaire grading code when the checkbox question is not answered correctly and the `partial_points` setting is disabled.

Take a look at how the existing note HINT_MULTIPLE_CHOICES_SELECTABLE has been added:
https://github.com/apluslms/mooc-grader/blob/a86071f57a1403e9650d7465b196af41a4b577f7/access/types/forms.py#L684

In the same `grade_field` function, you can add the new hint. You can use the same django.po messages you already added. The function already has access to the settings like `partial_points` and it knows if the checkbox question was answered correctly.

ihalaij1:
I made changes according to the requests you made. If multiple hints are displayed, the new hint always appears last.
Here's how it looks now:
image

Mankro (approved):
Looks great!

Mankro:
I changed the git commit message and force-pushed.

Mankro:
I now noticed that the original issue requested a slightly different trigger for the new hint:

> IMO, it would be a good idea to insert an additional feedback bullet on pick-anys, to be displayed whenever a student’s answer is partially correct (i.e., something was right, but not everything).

In this pull request, the hint is triggered even if the submission is not partially correct, but completely incorrect. However, I don't think this is an important distinction. After all, it is easy to make partially correct submissions since only one checkbox needs to be in the correct state in order to make it "partially" correct. Each checkbox is either ticked or not ticked. If the student leaves one easy obviously incorrect choice unticked, that already makes it a partially correct submission.

jsorva:
Looks good to me, thanks @ihalaij1 and @Mankro!

